Issue #5, opened 2019-08-20
Dockerize


n1try:

robrotheram:
Hi I have created the above PR to Dockerize the application

n1try:
Resolved with #8

PR #8, opened 2019-10-07
Dockerise The application 

robrotheram:
Hi, Happy hacktoberfest 
This PR address #5  with some added fixes:
* Typo in one of the models
* Updated to use go modules and go 1.13
* The Docker file use multistage so it will first compile the application and then create a smaller container with the binary from the first container
*  Also added a docker-compose file with mysql db container so that you can now run this application in one command :smile:


n1try:
Great contribution, thanks a lot! 👍 
Happy Hacktoberfest 🍻

PR #19, opened 2020-04-20
Feat/add sqlite support

### LINKED ISSUE #17: SQLite Support

will7200:
Closes #17 
Due to struct tags not being able to be modified at runtime, its quite difficult to work with different sql flavors with Gorm. 
Foreign keys in sqlite need to be specified when the table is created.
MySql doesn't allow the references next to the column definition.
To that end, I used a migration strategy to add support for sqlite and maintain current support with MySql and Postgres.


muety:
Hi @will7200, thanks a lot for your contribution! I'm going to have a deeper look at your PR later. Two things in advance:

1. Please re-check line indentation in the SQL script
2. Can't we use GORM's `AutoMigrate` with SQLite as well? I'd really try to avoid special cases like the one in `main.go`, unless they're totally unavoidable.

will7200:
Hey @muety 
1. What’s wrong with the indentation? Was formatted using jetbrains database tool?
2. Well

muety:
### SQL indentation
> What’s wrong with the indentation?

I don't really like the way how `primary key` and `autoincrement` modifiers are weirdly placed into the next line with odd indentation. 

```sql
id      integer
        primary key autoincrement,
type    integer      not null,
```

But it seems to be intended, so let's just keep it that way.

### Migrations and foreign keys
I see the problem. Looks like things are going to be a lot easier with Gorm v2 (https://github.com/jinzhu/gorm/issues/450), since it allows for defining foreign keys "inline" and without requiring a separate `ALTER TABLE` statement, which is not supported by SQLite. Unfortunately, v2 is still under development and migrating from v1 to v2 will probably be a non-trivial effort, anyway. So let's stick with your (very elaborate!) solution for now. Only two further points I'd like you to consider:

1. Could you maybe move the migration-related code in line 128 and following (https://github.com/muety/wakapi/pull/19/files#diff-7ddfb3e035b42cd70649cc33393fe32cR128) to a separate utils method or so? I'd like to keep `main()` as short and clean as possible. Maybe introduce a "factory"-method somewhere that returns the `migrationDo()` function, depending on the selected type of database.
2. Do we even need real "migrations"? Or can we just run the "up"-part once initially and be done? I'm asking because to have migrations introduces another two external dependencies.

muety:
@will7200 Any news on this?

will7200:
@muety 
1. I can move it to a separate function within the main file. I would like to maintain any database actions visible at the top level, just as it was before.
2. Yes. If we expect to add or change any data structures then that would inevitable mean changing our database structure. If we don't maintain migrations we would only be welcoming new users and leaving existing users to not able to use a newer version or forcing them to drop their existing data.

muety:
Alright, please just clean up the `main()` method a little and this can be merged. Thanks a lot for your contribution!

will7200:
@muety 
Done!

PR #32, opened 2020-05-30
Automatically build the project for Windows users

NickAcPT:
This change makes it simpler for Windows users to use the project by automatically building the project with GitHub Actions on every release.

This allows for an easier way to use the project by automatically adding a zip file with the built executables to new releases.

Whenever you add a new Release, this GitHub Action will start and will build for Windows.

Example release:
Example Release
As you can see, the built zip file is added automatically

### DIFF
--- a/.github/workflows/linux-build-on-release.yml
+++ b/.github/workflows/linux-build-on-release.yml
@@ -1,0 +1,43 @@
+name: Build WakAPI on Linux
+
+on:
+  release:
+    types: 
+      - created
+
+jobs:
+  build-and-release:
+    name: Build and add to Release
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+      
+    - name: Build
+      run: GO111MODULE=on go build -v .
+
+    - name: Zip Release
+      uses: TheDoctor0/zip-release@v0.3.0
+      with:
+        filename: release.zip
+
+    - name: Upload built executable to Release
+      uses: actions/upload-release-asset@v1.0.2
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ github.event.release.upload_url }}
+        asset_path: release.zip
+        asset_name: wakapi_linux.zip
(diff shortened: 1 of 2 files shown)


muety:
Great contribution, thanks a lot! I've never worked with GitHub actions before, so that's very interesting for me.

Would you mind adding a Linux amd64 build as well?

muety:
Is there a way to test this, except for merging and actually running the action through creating a new release?

I wonder if the build will actually be successful. As far as I know, you need MinGW on Windows to build the project (more specifically the `go-sqlite3` dependency), which is probably not inlcuded by default, or is it?

muety:
Never mind, I saw you already have a working release at https://github.com/LightPOS/wakapi/releases.

But could you please quickly explain how why the SQLite3 compilation works in that action?

NickAcPT:
> Great contribution, thanks a lot! I've never worked with GitHub actions before, so that's very interesting for me.

This is also kinda new to me.

> Would you mind adding a Linux amd64 build as well?

I could try, but I don't have much experience with doing Linux stuff.
All that is needed is to change some actions (enabling go 1.11 modules and zipping).

> But could you please quickly explain why the SQLite3 compilation works in that action?

According to [this](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md), we can see that they have pre-installed a lot of stuff to enable builds.
So, the 64 bits version is already installed and everything works fine.

NickAcPT:
@muety Okay, I added the Linux file and it builds and executes correctly.
Release example

We could probably do MacOS in the future, but it is waaay out of my league.

NickAcPT:
Ermm, there's something wrong.. The actions didn't run

muety:
Yes, I know. But I can't figure out the problem. Actions are enabled for the project and the workflow definitions are there. But no actions are being detected. Any ideas?

NickAcPT:
I'm assuming it is something broken because of the PR. Could you try moving the files out of the git repo (effectively deleting the files from git) and readding them?

NickAcPT:
[Here](https://github.community/t/no-github-actions-workflows-detected-in-repo/16085/12) someone mentioned to remove the .github folder and reverting the commit (which results in the same thing I suggested)

muety:
Didn't help. I'll wait if someone on that forum post has a solution.

NickAcPT:
Could you try doing it from the web interface? GitHub might only allow Actions from "Verified" commits only. Also, don't forget to create a new release do that it'll trigger the actions.

NickAcPT:
@muety by the way, did you publish the release directly or was it a draft?
Draft releases don't notify these workflows when they get published.

muety:
It was indeed a draft. I'll try an immediate release later.

PR #35, opened 2020-07-14
feat: add base_path environment variable

YC:
Allow modification of `base_path` through environment variable.
Happy to change `wait-for-it.sh` to modify `config.ini` instead also, if you're unhappy with the current approach.


muety:
Thanks! :slightly_smiling_face:

PR #62, opened 2020-10-25
Add user-managed custom rules feature

notarock:
Hey there. I've been using this project for a while now and thought it would be great to let users input their own custom rules to changes the heartbeat's language based on the filename. While searching the code, I found something very similar [here](https://github.com/muety/wakapi/blob/177cbb12fcc46f9f63b4f65b482847eaad6e5f98/config.default.yml#L11) and thought it would be great to have a way to edit those rules on a per-user basis.

To do this, I added a custom_rule model and services and added endpoints to let the user create/delete them. Then, when an heartbeat is received, the rule is applied in the `Augment` function.

## Heres some screenshot: 

I was inspired by this (From wakatime):

image

### In Wakapi :

image

image

image

Let me know what you think. BTW, are there any linter / code format tools I should be using for go? Should I use any of them before opening a PR?

### DIFF
--- a/routes/heartbeat.go
+++ b/routes/heartbeat.go
@@ -13,14 +13,16 @@ import (
 )
 
 type HeartbeatHandler struct {
-	config        *config2.Config
-	heartbeatSrvc *services.HeartbeatService
+	config         *config2.Config
+	heartbeatSrvc  *services.HeartbeatService
+	customRuleSrvc *services.CustomRuleService
 }
 
-func NewHeartbeatHandler(heartbeatService *services.HeartbeatService) *HeartbeatHandler {
+func NewHeartbeatHandler(heartbeatService *services.HeartbeatService, customRuleService *services.CustomRuleService) *HeartbeatHandler {
 	return &HeartbeatHandler{
 		config:        config2.Get(),
 		heartbeatSrvc: heartbeatService,
+		customRuleSrvc: customRuleService,
 	}
 }
 
@@ -41,13 +43,20 @@ func (h *HeartbeatHandler) ApiPost(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	rules, err := h.customRuleSrvc.GetCustomRuleForUser(user.ID)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte(err.Error()))
+		return
+	}
+
 	for _, hb := range heartbeats {
 		hb.OperatingSystem = opSys
 		hb.Editor = editor
 		hb.Machine = machineName
 		hb.User = user
 		hb.UserID = user.ID
-		hb.Augment(h.config.App.CustomLanguages)
+		hb.Augment(rules)
 
 		if !hb.Valid() {
 			w.WriteHeader(http.StatusBadRequest)
--- a/routes/heartbeat.go
+++ b/routes/heartbeat.go
@@ -56,7 +56,8 @@ func (h *HeartbeatHandler) ApiPost(w http.ResponseWriter, r *http.Request) {
 		hb.Machine = machineName
 		hb.User = user
 		hb.UserID = user.ID
-		hb.Augment(rules)
+		hb.AugmentWithConfigRules(h.config.App.CustomLanguages)
+		hb.AugmentWithUserRules(rules)
 
 		if !hb.Valid() {
 			w.WriteHeader(http.StatusBadRequest)
--- a/routes/settings.go
+++ b/routes/settings.go
@@ -116,7 +116,7 @@ func (h *SettingsHandler) DeleteCustomRule(w http.ResponseWriter, r *http.Reques
 	user := r.Context().Value(models.UserKey).(*models.User)
 	ruleId, err := strconv.Atoi(r.PostFormValue("ruleid"))
 	if err != nil {
-		respondAlert(w, "internal server error", "", "settings.tpl.html", http.StatusInternalServerError)
+		respondAlert(w, "internal server error", "", conf.SettingsTemplate, http.StatusInternalServerError)
 		return
 	}
 
@@ -149,7 +149,7 @@ func (h *SettingsHandler) PostCreateCustomRule(w http.ResponseWriter, r *http.Re
 	};
 
 	if _, err := h.customRuleSrvc.Create(rule); err != nil {
-		respondAlert(w, "internal server error", "", "settings.tpl.html", http.StatusInternalServerError)
+		respondAlert(w, "internal server error", "", conf.SettingsTemplate, http.StatusInternalServerError)
 		return
 	}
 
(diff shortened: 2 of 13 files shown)


muety:
Hi @notarock, thanks a lot for your contribution! I like it, however, I have a few remarks regarding your implementation, which I would like you to take a look at:

* [ ] Did you test with MySQL or Postgres? Because it looks like the new tables are not created in case of using one of those. The reason is probably, that you didn't specify to `AutoMigrate` the new model (see [here](https://github.com/muety/wakapi/blob/master/config/config.go#L95)). There's a (bit ugly) distinction between using SQLite and one of the other two databases. For MySQL and Postgres, table creation and modification is handled by GORM through its auto migration functionality. However, for SQLite, this doesn't work. So in the case of SQLite, database modifications are done using raw SQL scripts in the `migrations` directory, like the one you provided.
* [ ] Please take a look at `config/config.go` and `models/config.go` (added in your commit). There seems to be an overlap, but mostly, they're the same. There has been some refactoring to the config recently, so  I assume a merge conflict was not resolved properly here? 
* [ ] Although your implementation gives more flexibility to individual users, it essentially overrides the old functionality provided by the `custom_languages` config. Imagine you're hosting a Wakapi instance with several users and have defined `custom_languages` globally on your server. With your change, they're obsolete, i.e. from the moment the new version is deployed, `.vue` (for instance) will never be resolved to `Vue` unless each user adds a custom rule for themselves. I'd like to keep supporting the global custom languages and I see two options. *Option 1* is to keep both in parallel, the `custom_languages` and your new custom rules feature. Both are applied two new heartbeats, while user-defined custom rules take precedence over server-defined rules in case of conflicts. *Option 2* is to automatically migrate the "legacy", server-defined rules and add according custom rules for every currently registered user. Afterwards (some time in the future), the the `custom_languages` can be abolished entirely. This option removes any control about extension-to-language replacements from the server operator. But at least already existing behavior wouldn't be changed from one moment to the other. I like option 1 better, but it's probably also more effort to implement. What do you think?
* [ ] Please at a very brief (~ one sentence) explanation in the "Custom Rules" section of the UI about the new feature, as "custom rules" is not necessarily very descriptive by itself and a user wouldn't immediately know that they can specify extension-to-language mapping there.
* [ ] Minor: if you want, make the new input fields in the UI the same height (incl. padding) as the other input fields on that page and maybe also align them vertically. 
* [ ] Few more minor changes, which I'll comment right in the code.

services/custom_rule.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func NewCustomRuleService(db *gorm.DB) *CustomRuleService {
+	return &CustomRuleService{
+		Config: models.GetConfig(),
+		Db:     db,
+	}
+}
+
+func (srv *CustomRuleService) GetCustomRuleById(CustomRuleId uint) (*models.CustomRule, error) {
```
muety, line 20: Minor: please make parameter names lower case.

routes/settings.go
```diff
@@ -105,6 +108,56 @@ func (h *SettingsHandler) PostCredentials(w http.ResponseWriter, r *http.Request
...
+	}
+
+	rule := &models.CustomRule{
+		UserID: user.ID,
+		Extension: extension,
+		Language: language,
+	};
+
+	if _, err := h.customRuleSrvc.Create(rule); err != nil {
+		respondAlert(w, "internal server error", "", "settings.tpl.html", http.StatusInternalServerError)
```
muety, line 152: Minor: please use constant instead of hard-coded `"settings.tpl.html"`. See other route definitions for reference.

routes/heartbeat.go
```diff
@@ -41,13 +43,20 @@ func (h *HeartbeatHandler) ApiPost(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	rules, err := h.customRuleSrvc.GetCustomRuleForUser(user.ID)
```
muety, line 46: Every new heartbeat will cause a database request to fetch user rules. This seems a bit inefficient. We should use caching here. At some other part of the application, we are using `go-cache` anyway, so I'd be happy if we could utilize it here as well.

muety (changes requested):

muety:
Sorry, among all those change requests I want to emphasize again that I highly appreciate your efforts! I'm glad that there are people who help bringing Wakapi forward :slightly_smiling_face:

notarock:
Fair enough, I was prepared for code review because I was more focused on making this work than writing clean code at first :eyes: . As for the questions:

- I tested this using Postgres with the docker-compose setup. This means that the database was fresh everytime I ran the migrations. I will make sure that the SQLite migrations works when I go over your changes requests, because I'm not too sure if I tested it yet.
- `Config.go` changed while I was developing and I didn't constantly rebase on your master, so its probably just a rebase/merge conflict error...
- For the `custom_language` config, I think I will put back the old `Augment` function (Maybe rename it to `AugmentWithConfigRules`), and create a new `AugmentWithUserRules` that will works based on the user's rules.
- True. It's not very user-friendly if there's no way to know what it does.
- Will fix
- Will fix

routes/heartbeat.go
notarock: I changed the customRuleService to use the cache before doing a database request, this way we can easily remove cached rules when they are modified.

routes/heartbeat.go
```diff
@@ -41,13 +43,21 @@ func (h *HeartbeatHandler) ApiPost(w http.ResponseWriter, r *http.Request) {
...
+
 	for _, hb := range heartbeats {
 		hb.OperatingSystem = opSys
 		hb.Editor = editor
 		hb.Machine = machineName
 		hb.User = user
 		hb.UserID = user.ID
-		hb.Augment(h.config.App.CustomLanguages)
+		hb.AugmentWithConfigRules(h.config.App.CustomLanguages)
+		hb.AugmentWithUserRules(rules)
```
notarock, line 60: This keeps the old and new behavior of the Augment function.

routes/settings.go
notarock: Fixed

services/custom_rule.go
notarock: Fixed it

notarock:
I just tested with SQLite and everything works as expected.

notarock:
I fixed the style, it now looks like this:

image

image

LMK if it's missing something

muety:
Thanks a lot for respecting those change requests! I'll have a detailed look at it later, but on first sight everything looks great now.

muety:
Jfyi: I added [this little change](https://github.com/muety/wakapi/commit/e19761337fb2af366dc58c966a09c02016089e7f#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R102) in addition, which was required to properly set up tables in existing MySQL / Postres instances. Also, I did a few minor adaptions to the compose file.

Thanks again for your contribution :+1: Hope to see more from you in the future!

PR #77, opened 2020-12-01
Added missing closing parens in language mapping description

notarock:
Fix a very minor typo where the closing parens was missing.
## Before
image

## After
image


muety:
Thanks!

PR #85, opened 2021-01-12
Reduce minimum username length to 1

YC:
I would like to have a 2 character username. 
Feel free to close if you think that the minimum should be kept as is.


muety:
Thanks for the contribution! There's nothing against 2-character user names. Please increment the patch version in `version.txt`, afterwards I'm gonna merge this.

PR #95, opened 2021-01-22
feat: embed assets into binary

### LINKED ISSUE #26: Package assets into executable
It would be convenient to be able to package Wakapi as a single, executable binary that also contains non-Go files, i.e. e.g. `data/colors.json`, `views/*.html`, `static/*`, `version.txt`, etc. As far as I know, [go-bindata](https://github.com/go-bindata/go-bindata) is the way to go here, but I haven't worked with it, yet.

What should NOT be included in the binary is:
* `.env`
* `config.ini`
* `*.db`

YC:
- Embeds static files using pkger
- `go get github.com/markbates/pkger/cmd/pkger` and `go generate` needs to be used to generate `pkged.go`, not sure how to add the module into `go mod`
- If `pkged.go` is missing, it seems to look for the file locally

Resolves #26


muety:
Thanks a lot for this contribution, I appreciate it! I have a few remarks.

1. I might be missing something here, but for me, static assets don't seem to work anymore (see screenshot below). However, files like `version.txt` seem to be loaded correctly, as can be seen in the bottom left corner of the page. Looks like the routes to the static assets can't be resolved. I did `go generate`, then `go build -o wakapi` and then ran the binary. Btw, if I run the application without `pkged.go` being present, those `404` responses are turned into `500` ones. Could you take a look, please? Is it working for you?
2. I'm not entirely aware of how these packagers work, so I can't tell if the following is difficult to realize, if possible at all. However, I'd like to have everything work just like before during development and only have the packaging stuff kick in for production. Especially, I'd ike to be able to change the HTML templates and just refresh my browser without having to recompile and restart the whole application. Is it possible? 
3. I'd like to have things like HTML templates, static CSS, JS and images, etc. be packaged into the binary. However, `config.yml` should be read locally from the file system. That is, when deployed to a server, there should be nothing but `wakapi` (the binary) and a `config.yml`.
4. How and where to you specify what things to package in the first place? For instance, a local `wakapi.db` SQLite database file, that I accidentally have in my project root at build time, shouldn't get packaged.

Sorry for that wall of text! I hope this won't be too much effort for you. Maybe it's only that I did some stupid mistake while testing and some of the above points are already working anyway?

Screenshot from 2021-01-23 17-22-54

YC:
1. Sorry about this, don't know how I missed it. It is indeed not working. It will be fixed. 
2. When `pkged.go` is not present, it should use local files. I've just tested the template refresh scenario and it seems to work fine. 
3. `pkger list` should tell you the files which are included
4. See ### https://github.com/markbates/pkger#understanding-the-parser
Just discovered this so I will remove the -include arguments at the top of `main.go` in my upcoming commit.

muety:
Very cool, works exactly the way I wanted now! Thanks!

muety:
Ah btw, are you familiar with GitHub actions by any chance? If yes, would you mind adapting the releases actions for Linux and Windows (see [here](https://github.com/muety/wakapi/blob/master/.github/workflows/linux-build-on-release.yml)) in a way that the ZIP artifact only contains the binary and the config file, but not all the source files, assets, etc.? No worries if not! Then I'll give it a try myself tonight.

YC:
Sure, I'll give it a try

muety:
Thanks a lot!

You'll also need to adapt the workflows to include the two additional build steps:
* `go get github.com/markbates/pkger/cmd/pkger` and
* `go generate`

The action used for zipping the stuff is [zip-release](https://github.com/TheDoctor0/zip-release). However, it looks like you can only specify patterns to exclude, but not to include. 

I was thinking of first creating a subfolder like `dist`, copying `config.default.yml` into there, also putting the binary their (`go build -o dist/wakapi`) and then zipping only that subfolder.

Let me know if you need any assistance and please don't feel pressured to do this :)

muety:
You might be able use [act](https://github.com/nektos/act) to test GitHub actions locally, but I haven't tried it, yet.

PR #107, opened 2021-01-30
chore: remove legacy config.ini and .env

### LINKED ISSUE #106: Get rid of legacy ini config
A while ago, a YAML based config format was introduced alongside the ability to override variables via environment. It is powered by [configor](https://github.com/jinzhu/configor). A migration function was written to still support `config.ini` for compatibility reasons. However, in the mean time, most users should have been migrated to the new format, so all logic related to the old config format might be thrown out.

YC:
Resolves #106


muety:
Thanks a lot! I'll test and review it later.

PR #140, opened 2021-03-06
Prevent invalid dates ranges in summary page

notarock:
It is currently possible to enter a "End" date that is before the startdate, or a "start" date that is after the end date. This should prevent the user from directly entering an invalid date by setting a "min" and "max" value on those two date pickers.

There are no validation or the server's side, but that shouldn't be a problem since the invalid date will not create an error, it will simply return no data.

I've added a simple function that update the min and max values when the other input's value changes. I'm not sure if this is the cleanest way to go about this :thinking:


muety:
Good catch, thanks a lot!

Issue #144, opened 2021-03-23
Custom Mappings: file extensions can't contain dots


tldev-de:
Hi there,

just found out, that custom mappings don't work when your file extension contains a dot.

![image](https://user-images.githubusercontent.com/11870921/112218294-d8901f80-8c23-11eb-9b59-35646f5155c7.png)

I use Laravel for a project and the blade templates are recognised as "XML" files instead of blade files (as I configured with my custom mapping).

PR #147, opened 2021-03-24
allow custom mappings with dot (issue #144)

tldev-de:
I had troubles to get my `.blade.php` files recognised as `Blade` files. Therefore I changed the mapping algorithm to check for a suffix instead of using a regex to get the extension (after the last dot in the filename) and compare this extension to the custom mappings.

I think this is a more convenient way for any use cases and won't break anything.

### DIFF
--- a/models/heartbeat.go
+++ b/models/heartbeat.go
@@ -4,15 +4,9 @@ import (
 	"fmt"
 	"github.com/emvi/logbuch"
 	"github.com/mitchellh/hashstructure/v2"
-	"regexp"
+	"strings"
 	"time"
 )
-
-var languageRegex *regexp.Regexp
-
-func init() {
-	languageRegex = regexp.MustCompile(`^.+\.(.+)$`)
-}
 
 type Heartbeat struct {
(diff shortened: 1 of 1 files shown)


muety:
Great contribution, thanks a lot!

models/heartbeat.go
```diff
@@ -40,15 +34,12 @@ func (h *Heartbeat) Valid() bool {
...
 func (h *Heartbeat) Augment(languageMappings map[string]string) {
-	groups := languageRegex.FindAllStringSubmatch(h.Entity, -1)
-	if len(groups) == 0 || len(groups[0]) != 2 {
-		return
-	}
-	ending := groups[0][1]
-	if _, ok := languageMappings[ending]; !ok {
-		return
+	for ending, value := range languageMappings {
+		if strings.HasSuffix(h.Entity, "."+ending) {
```
muety, line 38: Brief side note here: you could have used `fmt.Sptinf()`, which looks a bit cleaner to me, but might actually be slower, not sure.

Issue #177, opened 2021-04-18
Migration error >= 1.26.4


YC:
Database Engine: sqlite

On 1.26.4
```
2021-04-18T03:36:50.879195765Z [INFO ] potentially running migration '20201103-rename_language_mappings_table'
2021-04-18T03:36:50.907514853Z [INFO ] potentially running migration '20201106-migration_cascade_constraints'
2021-04-18T03:36:50.907545161Z [INFO ] not attempting to drop and regenerate constraints on sqlite
2021-04-18T03:36:50.907548256Z [INFO ] potentially running migration '20210202-fix_cascade_for_alias_user_constraint'
2021-04-18T03:36:50.907551532Z [INFO ] not attempting to drop and regenerate constraints on sqlite
2021-04-18T03:36:50.943274498Z [ERROR] table heartbeats already exists
panic: table heartbeats already exists

goroutine 1 [running]:
github.com/emvi/logbuch.(*StandardFormatter).Pnc(0xc0000a60a8, 0xc0001ff640, 0x1f, 0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/emvi/logbuch@v1.1.1/standard_formatter.go:57 +0xc5
github.com/emvi/logbuch.(*Logger).Fatal(0xc0000f2000, 0xc0001ff640, 0x1f, 0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/emvi/logbuch@v1.1.1/logger.go:132 +0xb3
github.com/emvi/logbuch.Fatal(...)
	/go/pkg/mod/github.com/emvi/logbuch@v1.1.1/funcs.go:55
github.com/muety/wakapi/migrations.RunSchemaMigrations(0xc0001e9a10, 0xc0000c9680)
	/src/migrations/migrations.go:39 +0x9e
github.com/muety/wakapi/migrations.Run(0xc0001e9a10, 0xc0000c9680)
	/src/migrations/migrations.go:33 +0x4d
main.main()
	/src/main.go:130 +0x389
```

On 1.26.3
```
2021-04-18T03:35:35.216092946Z [INFO ] potentially running migration '20201103-rename_language_mappings_table'
2021-04-18T03:35:35.217842192Z [INFO ] potentially running migration '20201106-migration_cascade_constraints'
2021-04-18T03:35:35.217855567Z [INFO ] not attempting to drop and regenerate constraints on sqlite
[...]

PR #181, opened 2021-04-22
sqlitemigrate patches for larger datasets

ThoreKr:
As a followup to #177 I fiddled with this script to migrate the data into postgres.

The way the postgres migration is currently fiddled in could probably be improved or dropped entirely since it adds additional obstacles (see point 4. -- for some reason the heartbeat id sequence was way off and caused errors on inserts on first start after the migration, not sure this holds for mysql as well).

Similar thing with the hash computation, not sure whether this was just an issue with my dataset or sqlite in general.

This 100 batching is however necessary, since the attempt to batch insert the entire dataset (~39k heartbeats in my case) fails for mysql and postgres.
For the former it complains about to many replacements in prepared statement, postgres complains about the exceeded number of parameters.

Arguably this implementation is a clunky hack. There is a more elegant way to do this, however, I haven't tested whether this has other side effects in wakapi itself: Use gorms `CreateBatch` here https://github.com/muety/wakapi/blob/8b9a9a1a42e420b93a43af47a65068748ae23a23/repositories/heartbeart.go#L32

### DIFF
--- a/scripts/sqlite2mysql.go
+++ b/scripts/sqlite2mysql.go
@@ -4,10 +4,11 @@ package main
 Usage:
 ---
 1. Set up a MySQL instance (see docker_mysql.sh for example)
-2. Create config file (e.g. migrate.yml) as shown below
-3. go run sqlite2mysql.go -config migrate.yml
+2. Create config file (e.g. config.yml) as shown below
+3. go run sqlite2mysql.go -config config.yml
+4. For postgres check https://wiki.postgresql.org/wiki/Fixing_Sequences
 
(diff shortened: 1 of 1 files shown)


muety:
Thanks for your contribution! Only did a few more tweaks.

I tried your script with both an empty MySQL and Postgres database and it worked just fine. However, the source SQLite database was not actually "real" data, but only a few thousand artificially generated heartbeats (via `sample_data.py`).

ThoreKr:
Sounds good. Thank you.

PR #213, opened 2021-06-23
feat: 🎸 add support for using a UNIX domain socket

Tatsh:
Can be tested with Netcat, socat, or [this code](https://gist.github.com/teknoraver/5ffacb8757330715bcbcc90e6d46ac74#file-unixhttpc-go).


muety:
Awesome work, thanks a lot! Just out of curiosity, in what situation would you want to prefer Unix sockets over IP?

Tatsh:
> Awesome work, thanks a lot! Just out of curiosity, in what situation would you want to prefer Unix sockets over IP?

This is how I run all my services that hook up to nginx. It's mainly for completeness and consistency.

PR #220, opened 2021-07-10
fix: remove unix socket if exists

YC:
@Tatsh can you please check?
I believe it is necessary to avoid `bind: address already in use`.


Tatsh:
Okay by me.

Tatsh (approved):

PR #229, opened 2021-08-19
Fix compatible with new wakatime-cli

sdvcrx:
Hi, the version number of new [wakatime-cli](https://github.com/wakatime/wakatime-cli) (golang version) starts with `v`, like [v1.18.11](https://github.com/wakatime/wakatime-cli/releases/tag/v1.18.11).

So this regular expression needs to be updated to support it.


muety:
Thanks a lot, very attentive!

Issue #235, opened 2021-08-29
Upgrade to Chart.js 3.x


muety:
We are currently using 2.9.4, which is the latest 2.x version. To benefit from the very latest changes to the library, we should consider upgrading to 3.x. However, I assume there have been breaking changes, so the upgrade will probably take non-trivial refactoring effort.

mawoka-myblock:
I'm having a look at the migration and I'll note which things have to be changed/which I have changed:
- [x] default colors: https://www.chartjs.org/docs/latest/getting-started/v3-migration.html#defaults
- [ ] Chart-tpye removed: https://www.chartjs.org/docs/latest/getting-started/v3-migration.html#defaults

muety:
Done with #236

PR #236, opened 2021-08-29
Upgraded to ChartsJs 3.x

mawoka-myblock:
With that, I addressed #235. How you see, there is still an error in the console, but I didn't noticed any visual bugs produced by this error.


mawoka-myblock:
@muety I think I won't go on with fixing the error in the console, so you can do it or you don't. Btw: thanks for your great work ❤️ !

muety:
Thanks for your efforts! I'll be on holidays without access to my computer the next couple of days, but will do the review afterwards.

With regard to Tailwind, please see https://github.com/muety/wakapi#tailwindcss.

YC:
> With that, I addressed #235. How you see, there is still an error in the console, but I didn't noticed any visual bugs produced by this error.

I traced this to:
```js
charts.forEach(c => c.options.onResize(c.chart))
```
Which works in the current version but not after upgrade. 
Removing `.chart` does seem to get rid of the error. 

Something else that I've noticed is that time has become an integer, rather than a formatted time.

mawoka-myblock:
> Thanks for your efforts! I'll be on holidays without access to my computer the next couple of days, but will do the review afterwards.
> 
> With regard to Tailwind, please see https://github.com/muety/wakapi#tailwindcss.

What do you mean?

mawoka-myblock:
> > With that, I addressed #235. How you see, there is still an error in the console, but I didn't noticed any visual bugs produced by this error.
> 
> I traced this to:
> 
> ```js
> charts.forEach(c => c.options.onResize(c.chart))
> ```
> 
> Which works in the current version but not after upgrade.
> Removing `.chart` does seem to get rid of the error.
> 
> Something else that I've noticed is that time has become an integer, rather than a formatted time.

For the integer-problem, I'll try my best but I don't promise anything, because my JavaScript skills aren't very good and I am not familiar with ChartsJs at all

YC:
I think I managed to fix the label issues. 
However, there is something wrong with toggling 'Show:' sometimes.

```
Uncaught Error: Canvas is already in use. Chart with ID '3' must be destroyed before the canvas can be reused.
    oo http://localhost:3000/assets/vendor/Chart.bundle.min.js:13
    draw http://localhost:3000/assets/app.js:210
    <anonymous> http://localhost:3000/assets/app.js:449
Chart.bundle.min.js:13:88709
```

mawoka-myblock:
> I think I managed to fix the label issues.
> However, there is something wrong with toggling 'Show:' sometimes.
> 
> ```
> Uncaught Error: Canvas is already in use. Chart with ID '3' must be destroyed before the canvas can be reused.
>     oo http://localhost:3000/assets/vendor/Chart.bundle.min.js:13
>     draw http://localhost:3000/assets/app.js:210
>     <anonymous> http://localhost:3000/assets/app.js:449
> Chart.bundle.min.js:13:88709
> ```

Yes, the problem occurs on the pie-diagrams every time I try to change the `Count` which also doesn't work

PR #238, opened 2021-08-30
Normalize information received in hearbeats

### LINKED ISSUE #234: Making the item name consistant between clients
**Describe the bug**
As you see in the screenshot, wakapi treat `windows` and `Windows` as different platform, also language `markdown` and `Markdown`, is there any way to fix this? i.e. do a transformation to upper/lower/title case for the item.
image

**System information**
* v1.17.2 @ postgres
* Operating system: docker n1try/wakapi:1.17.2
* Database (SQLite, MySQL, ... ?): postgres

notarock:
This pull requests add a `Normalize` function which transform some information in the heartbeat into a lowercase but titled string. Heartbeats are normalized after rules were applied in order to not create any conflict with them. 

This is an attempt to make Wakapi case insensitive, and should closes #234

I've had the same issue in the past (Ignore the `unknown` entries:)

image

 Now after using the Normalize functions, data looks like this:

image

I think this is a pretty simple solution, but I am concerned as to how this would behave with older information that were already saved from past heartbeats...

Do you have any suggestion on how we could address this?

### DIFF
--- a/models/heartbeat.go
+++ b/models/heartbeat.go
@@ -34,6 +34,12 @@ func (h *Heartbeat) Valid() bool {
 	return h.User != nil && h.UserID != "" && h.User.ID == h.UserID && h.Time != CustomTime(time.Time{})
 }
 
+func (h *Heartbeat) Normalize() {
+	h.Language = strings.Title(strings.ToLower(h.Language))
+	h.OperatingSystem = strings.Title(strings.ToLower(h.OperatingSystem))
+	h.Editor = strings.Title(strings.ToLower(h.Editor))
+}
+
 func (h *Heartbeat) Augment(languageMappings map[string]string) {
 	maxPrec := -1 // precision / mapping complexity -> more concrete ones shall take precedence
 	for ending, value := range languageMappings {
--- a/services/heartbeat.go
+++ b/services/heartbeat.go
@@ -179,6 +179,7 @@ func (srv *HeartbeatService) augmented(heartbeats []*models.Heartbeat, userId st
 
 	for i := range heartbeats {
 		heartbeats[i].Augment(languageMapping)
+		heartbeats[i].Normalize()
 	}
 
 	return heartbeats, nil


mawoka-myblock:
I don't like it, because I think it looks much prettier, if the stuff is Capitalized, so I'd suggest a parameter in the config-file or sth like this, so you can switch it off/on easily.

muety:
I'll review this some time in the next couple of days. Personally, I like OS, editor and language capitalized, but projects, for instance, shall be named just as they are in your IDE, which makes things a bit tricky. Can't think of a nice solution for that right now.

notarock:
> I'll review this some time in the next couple of days.

No worries

> Personally, I like OS, editor and language capitalized, but projects, for instance, shall be named just as they are in your IDE, which makes things a bit tricky. Can't think of a nice solution for that right now.

I had the same idea, so I only do it on Editor, Language, Os so far. Maybe this behavior should be configurable?  Since the project name are read from git, I don't think they should be touched in any way.

> I'd suggest a parameter in the config-file or sth like this, so you can switch it off/on easily. 

I didn't thought of this, but I like the idea. Maybe a simple per-user setting that enable or disable "Normalizing" those fields would do the job?

muety:
> I didn't thought of this, but I like the idea. Maybe a simple per-user setting that enable or disable "Normalizing" those fields would do the job?

Don't make it too complex, let's have at max a config flag. And normalizing OS, language, OS and editor, but not machines and projects sounds good to me for now.

muety:
I reviewed your code and everything looks fine so far, thanks! However, now that I thought about the problem a bit more deeply, I don't quite like the overall approach that much anymore, mainly because languages like "PHP" or "SQL" will end up as "Php" and "Sql" and "VSCode" will be "Vscode", which looks wrong to me. Unfortunately, I can't think of a more clever solution to this issue. Maybe we just stick to the user having to manually add an alias mapping?

mawoka-myblock:
> I reviewed your code and everything looks fine so far, thanks! However, now that I thought about the problem a bit more deeply, I don't quite like the overall approach that much anymore, mainly because languages like "PHP" or "SQL" will end up as "Php" and "Sql" and "VSCode" will be "Vscode", which looks wrong to me. Unfortunately, I can't think of a more clever solution to this issue. Maybe we just stick to the user having to manually add an alias mapping?

We could do it ONLY for Windows because I understood it that this problem only occurs on/for Windows.

muety:
What do you mean, "only for Windows"?

mawoka-myblock:
> What do you mean, "only for Windows"?

In this case, `windows` and `Windows` should be one thing, `Windows`, because the real name is spelled with a capital 'w'.

muety:
No, it happens for other OS' and editors as well.

notarock:
> I reviewed your code and everything looks fine so far, thanks! However, now that I thought about the problem a bit more deeply, I don't quite like the overall approach that much anymore, mainly because languages like "PHP" or "SQL" will end up as "Php" and "Sql" and "VSCode" will be "Vscode", which looks wrong to me. Unfortunately, I can't think of a more clever solution to this issue. Maybe we just stick to the user having to manually add an alias mapping?

This makes a lot of sense. I didn't realize there were names which would end up "messed up" when doing this - and in that case, it would make more sense to respect what wakatime plugins are sending in the heartbeats.

Issue #244, opened 2021-09-03
Implement statusbar endpoint


muety:
Implement endpoint that mimics https://wakatime.com/api/v1/users/current/statusbar/today.

**Example:**
```json
{
    "cached_at": "2021-09-03T23:19:10Z",
    "data": {
        "categories": [
            {
                "decimal": "1.00",
                "digital": "1:00:10",
                "hours": 1,
                "minutes": 0,
                "name": "Coding",
                "percent": 100.0,
                "seconds": 10,
                "text": "1 hr",
                "total_seconds": 3610.841188
            }
        ],
        "dependencies": [],
        "editors": [
            {
                "decimal": "1.00",
                "digital": "1:00:10",
                "hours": 1,
                "minutes": 0,
                "name": "Generator",
                "percent": 100.0,
                "seconds": 10,
                "text": "1 hr",
                "total_seconds": 3610.841188
            }
        ],
        "grand_total": {
            "decimal": "1.00",
            "digital": "1:00",
            "hours": 1,
            "minutes": 0,
            "text": "1 hr",
            "total_seconds": 3610.841188
        },
        "languages": [
            {
                "decimal": "0.30",
                "digital": "0:18:41",
                "hours": 0,
                "minutes": 18,
                "name": "JavaScript",
                "percent": 31.07,
                "seconds": 41,
                "text": "18 mins",
                "total_seconds": 1121.804656
            },
            {
                "decimal": "0.23",
                "digital": "0:14:13",
                "hours": 0,
                "minutes": 14,
                "name": "PHP",
                "percent": 23.64,
                "seconds": 13,
                "text": "14 mins",
                "total_seconds": 853.605674
            },
            {
                "decimal": "0.22",
                "digital": "0:13:49",
                "hours": 0,
                "minutes": 13,
                "name": "Blade Template",
                "percent": 22.97,
                "seconds": 49,
                "text": "13 mins",
                "total_seconds": 829.236203
            },
            {
                "decimal": "0.10",
                "digital": "0:06:07",
                "hours": 0,
                "minutes": 6,
                "name": "Python",
                "percent": 10.18,
                "seconds": 7,
                "text": "6 mins",
                "total_seconds": 367.500889
            },
            {
                "decimal": "0.08",
                "digital": "0:05:37",
                "hours": 0,
                "minutes": 5,
                "name": "Go",
                "percent": 9.34,
                "seconds": 37,
                "text": "5 mins",
                "total_seconds": 337.318179
            },
            {
[...]

PR #247, opened 2021-09-06
perf(dist): load external swagger-ui files

YC:
Loads swagger-ui files from CDN instead of embed. 
Reduces binary size from 42.4 -> 23.9MB. 
What do you think @muety?


muety:
Great contribution, thanks! Didn't know Swagger UI was that big. I'll merge this and publish a new release, soon. Wanna implement #244 first.

muety:
It only affects the Swagger UI, which are probably very rarely used anyway, not the web app itself. I think it's fine this way.

PR #248, opened 2021-09-06
Add statusbar endpoint (issue #244)

kondr1:
I really wanted this feature, so i made it =)
I am not sure about project structure, I made the handler in right package?
Waka docs don't have any information about this endpoint, but I think it needs for this project in another PR

### DIFF
--- a/routes/compat/wakatime/v1/statusbar.go
+++ b/routes/compat/wakatime/v1/statusbar.go
@@ -41,6 +41,7 @@ func (h *StatusBarHandler) RegisterRoutes(router *mux.Router) {
 	)
 	r.Path("/users/{user}/statusbar/{range}").Methods(http.MethodGet).HandlerFunc(h.Get)
 	r.Path("/v1/users/{user}/statusbar/{range}").Methods(http.MethodGet).HandlerFunc(h.Get)
+	r.Path("/compat/wakatime/v1/users/{user}/statusbar/{range}").Methods(http.MethodGet).HandlerFunc(h.Get)
 }
 
 // @Summary Retrieve summary for statusbar
@@ -61,7 +62,7 @@ func (h *StatusBarHandler) Get(w http.ResponseWriter, r *http.Request) {
 
 	rangeParam := vars["range"]
 	if rangeParam == "" {
-		rangeParam = (*models.IntervalPast7Days)[0]
+		rangeParam = (*models.IntervalToday)[0]
 	}
 
 	err, rangeFrom, rangeTo := utils.ResolveIntervalRawTZ(rangeParam, user.TZ())
@@ -86,8 +87,8 @@ func (h *StatusBarHandler) Get(w http.ResponseWriter, r *http.Request) {
 
 func (h *StatusBarHandler) loadUserSummary(user *models.User, start, end time.Time) (*models.Summary, int, error) {
 	summaryParams := &models.SummaryParams{
-		From:      time.Time{},
-		To:        time.Now(),
+		From:      start,
+		To:        end,
 		User:      user,
 		Recompute: false,
 	}
--- a/routes/compat/wakatime/v1/statusbar.go
+++ b/routes/compat/wakatime/v1/statusbar.go
@@ -1,0 +1,106 @@
+package v1
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/gorilla/mux"
+	conf "github.com/muety/wakapi/config"
+	"github.com/muety/wakapi/middlewares"
+	"github.com/muety/wakapi/models"
+	v1 "github.com/muety/wakapi/models/compat/wakatime/v1"
+	routeutils "github.com/muety/wakapi/routes/utils"
+	"github.com/muety/wakapi/services"
+	"github.com/muety/wakapi/utils"
+)
+
+type StatusBarViewModel struct {
+	CachedAt time.Time        `json:"cached_at"`
+	Data     v1.SummariesData `json:"data"`
+}
+
+type StatusBarHandler struct {
+	config      *conf.Config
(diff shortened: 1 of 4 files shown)


muety:
Thanks a lot! Looks good. Only a few remarks here (see code comments).

routes/compat/wakatime/v1/statusbar.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (h *StatusBarHandler) RegisterRoutes(router *mux.Router) {
+	r := router.PathPrefix("").Subrouter()
+
+	r.Use(
+		middlewares.NewAuthenticateMiddleware(h.userSrvc).Handler,
+	)
+	r.Path("/users/{user}/statusbar/{range}").Methods(http.MethodGet).HandlerFunc(h.Get)
+	r.Path("/v1/users/{user}/statusbar/{range}").Methods(http.MethodGet).HandlerFunc(h.Get)
```
muety, line 43: For the sake of consistency, please also add a "compat" route, analogously to this one: https://github.com/muety/wakapi/blob/2f9cafc88c10b5b1384997232418c30d82f04281/routes/compat/wakatime/v1/stats.go#L35

routes/compat/wakatime/v1/statusbar.go
```diff
@@ -0,0 +1,106 @@
...
+func (h *StatusBarHandler) Get(w http.ResponseWriter, r *http.Request) {
+	user, err := routeutils.CheckEffectiveUser(w, r, h.userSrvc, "current")
+	if err != nil {
+		return // response was already sent by util function
+	}
+	var vars = mux.Vars(r)
+
+	rangeParam := vars["range"]
+	if rangeParam == "" {
+		rangeParam = (*models.IntervalPast7Days)[0]
```
muety, line 64: Make `today` the default, as this endpoint will almost always be used to retrieve today's stats.

routes/compat/wakatime/v1/statusbar.go
```diff
@@ -0,0 +1,106 @@
...
+	summariesView := v1.NewSummariesFrom([]*models.Summary{summary}, &models.Filters{})
+	utils.RespondJSON(w, r, http.StatusOK, StatusBarViewModel{
+		CachedAt: time.Now(),
+		Data:     *summariesView.Data[0],
+	})
+}
+
+func (h *StatusBarHandler) loadUserSummary(user *models.User, start, end time.Time) (*models.Summary, int, error) {
+	summaryParams := &models.SummaryParams{
+		From:      time.Time{},
```
muety, line 89: You're ignoring `start` and `end` here, instead, you're always returning the user's entire coding time.

routes/compat/wakatime/v1/statusbar.go
```diff
@@ -0,0 +1,106 @@
...
+		retrieveSummary = h.summarySrvc.Summarize
+	}
+
+	summary, err := h.summarySrvc.Aliased(summaryParams.From, summaryParams.To, summaryParams.User, retrieveSummary, summaryParams.Recompute)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	return summary, http.StatusOK, nil
+}
```
muety, line 106: In addition, it would be could if you added two or three basic API tests to the corresponding Postman collection (see https://github.com/muety/wakapi/tree/master/testing). You can take the _"Get stats ..."_ requests as an example. Shouldn't take long after you initially set up your Postman. Further information about our API tests can be found at https://github.com/muety/wakapi#api-tests.

muety (changes requested):

muety:
I just noticed, that this endpoint is almost the exact same as the [summaries endpoint](https://wakatime.com/developers#summaries). [Here](https://github.com/wakatime/wakatime-cli/blob/4c661de6441a35dd4d81f9e9d5337e0c2109cc81/pkg/api/summary.go#L17) is CLI's code for performing the statusbar request. Instead of more or less duplicating most of the code, can't we just return a `307` here to simply redirect to `/compat/wakatime/v1/users/{user}/summaries?range={range}`? I think that'd make an even better solution. What do you think?

kondr1:
> I just noticed, that this endpoint is almost the exact same as the [summaries endpoint](https://wakatime.com/developers#summaries). Instead of more or less duplicating most of the code, can't we just return a `307` 

Yes, it would be good solution, I think. 🤔

routes/compat/wakatime/v1/statusbar.go
kondr1: In case with redirect test needed?

kondr1:
> > I just noticed, that this endpoint is almost the exact same as the [summaries endpoint](https://wakatime.com/developers#summaries). Instead of more or less duplicating most of the code, can't we just return a `307`
> 
> Yes, it would be good solution, I think. 🤔

No, that's not working =( In summaries endpoint "data" is array but in status bar is not

muety:
> No, that's not working =( In the summaries endpoint "data" is array but in statusbar is not

My fault. It's fine then.

Thanks for your changes! Only one more thing. There are two Postman collections, one just to make making requests more convenient for the developer and another one that actually contains API tests (see my above comment). I appreciate you added the statusbar request to the first one. In addition, I'd like to have a few very basic tests cases in the second collection. Just verify that a proper status code is returned and that the schema of the response body's content is roughly correct.

Here is an example for the stats endpoint and in the left pane you can see all other test cases. Please just add some, which you consider meaningful, more the statusbar endpoint and then this thing is ready to me merged :muscle: 
image

Again, thanks a lot for your contribution.

muety (approved):

kondr1:
Oh, ok. I didn't see this collection.

PR #252, opened 2021-09-08
Config for working with repo use gitpod

kondr1:


muety:
I've never worked with GitPod before, but don't you need `init` and `command` hooks as well (https://www.gitpod.io/docs/config-start-tasks/)?

kondr1:
No, these tasks are not required.

PR #271, opened 2021-11-25
README cleanup

YC:


muety:
Thanks for cleaning up! Why did you remove the ToC, though? I kind of liked it.

YC:
Github has the menu already (top left), and it didn't seem to work on Docker Hub. 
Happy to revert if you prefer to have it still.

YC:
Oh, I see that this menu is more condensed. 
Will revert.

muety:
I actually never realized that built-in table contents before, thanks for pointing me there 😄. In that case I'd actually want to remove the one in the README, so feel free to revert your revert again and then just merge yourself :+1:

PR #279, opened 2021-12-14
fix: anticipated docker push issue

### LINKED ISSUE #278: Unable to publish Docker image
See https://github.com/muety/wakapi/runs/4518181679?check_suite_focus=true.

@YC Could you please have a look what's wrong here? Maybe `settings.GITHUB_TOKEN` is missing? 

image

YC:
Resolves #278

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -33,33 +33,29 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Extract metadata (tags, labels) for Docker
-        id: meta
-        uses: docker/metadata-action@v3
-        with:
-          images: |
-            n1try/wakapi
-            ghcr.io/${{ github.repository }}
-
-      - name: Build and push to Docker Hub
+      - name: Build and push
         uses: docker/build-push-action@v2
         with:
           push: true
           tags: |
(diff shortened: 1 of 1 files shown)


muety:
Thanks your super fast reaction! Can you give a brief explanation, please?

YC:
> Thanks your super fast reaction! Can you give a brief explanation, please?

I encountered this issue around 2 weeks ago, when I wanted to push one of my docker image to ghcr with a slightly adjusted version of our workflow. 

This is an example run of our last working configuration.
https://github.com/muety/wakapi/actions/runs/1328392599/workflow
This PR adds ghcr tags on top of that. 

I don't believe we can use metadata-action in the intended way since we are publishing 2 images (alpine and non-alpine). 
https://github.com/docker/metadata-action#basic
https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages
However, this does show that it's possible to publish to both Docker Hub and Github Packages with one workflow and I'm pretty certain that we don't need to provide the Github credentials explicitly. 

I haven't tried it out in practice, but here is a tutorial with this configuration: 
https://event-driven.io/en/how_to_buid_and_push_docker_image_with_github_actions/

muety:
Thanks a lot for that fix and your detailed explanation! :)

Issue #266, opened 2021-10-25
Support auto-refresh page in the frontend


gaocegege:
The page is static and the info may be outdated. It is better to support auto-refresh. For example, refresh the page/data on the page every 5 minutes

muety:
See https://github.com/muety/wakapi/pull/304#issuecomment-1007943696 for some thought on implementation.

PR #304, opened 2022-01-07
feat: auto-refresh

mawoka-myblock:
Added auto-refresh every 300 seconds (5 mins) for the summary-page. Compatibility: unkown, but works on latest chromium. It works completely without javascript. "Fixes" #266

### DIFF
--- a/views/summary.tpl.html
+++ b/views/summary.tpl.html
@@ -5,6 +5,7 @@
 
 <script src="assets/js/components/time-picker.js"></script>
 <script type="module" src="assets/js/components/summary.js"></script>
+<meta http-equiv="refresh" content="300"/>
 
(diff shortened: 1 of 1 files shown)


muety:
Thanks for your contribution. However, to be honest, I find the page regularly refreshing itself - potentially while I'm currently inspecting some charts - more annoying than manually hitting F5 once.

mawoka-myblock:
> 
Maybe a switch per user would be good!

muety:
I'd rather want to implement it properly. That might include having an event stream (probably using SSE) between server and website and then partially refreshing `wakapiData.*` each time a heartbeat arrives. However, this change is far from trivial, as we would have to implement parts of the heartbeat / duration aggregation logic in the frontend and change the charts' code structure so they can be updated / re-rendered dynamically. This will probably take me a day or too. I'll have a look when I get to it. Thanks a lot for your efforts, but I'll drop this PR for now.

PR #308, opened 2022-01-16
ci: major and major.minor tags for Docker publish

### LINKED ISSUE #307: Tag images with major versions
I find it's best when using containers to tag the image to a specific release, mostly the major release. That way people choose when to upgrade their instance, and can read the changelogs etc.

Right now we only have the exact version and the latest tag, so I propose tagging new images with the major version as well.

YC:
Resolves #307

@muety Please also enable packages on home page

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,10 +10,6 @@ jobs:
   docker-publish:
     runs-on: ubuntu-latest
     steps:
-      # https://stackoverflow.com/questions/58177786
-      - name: Get version
-        run: echo "GIT_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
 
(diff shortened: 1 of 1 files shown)


muety:
Great work again!

muety:
Haven't really used GitHub's Packages, yet, but it looks super convenient. I'm a big fan of having everything in one place, so I'm even thinking of ceasing to publish to DockerHub in the long run. What do you think would be a good "migration path"? Do an announcement in the discussions instructing people to switch to GHCR and eventually deleting the DockerHub repo after sufficiently long time?

YC:
> Haven't really used GitHub's Packages, yet, but it looks super convenient. I'm a big fan of having everything in one place, so I'm even thinking of ceasing to publish to DockerHub in the long run. What do you think would be a good "migration path"? Do an announcement in the discussions instructing people to switch to GHCR and eventually deleting the DockerHub repo after sufficiently long time?

Personally, I'd be inclined to keep Docker Hub, as it's the default with it comes to docker cli and hosts all the major base and app images. Is there much overhead in maintaining Docker Hub?

PR #317, opened 2022-01-29
Implement wakatime compatible /heartbeats endpoint

### LINKED ISSUE #241: Expose raw heartbeats
Create an endpoint analogously to WakaTime's one to expose raw heartbeats data.

YC:
Resolves #241

~~Possible issues with PR:~~
~~- Format of `date` (it's not clearly specified what format it should be in)~~
~~- More routes should be implemented (e.g. /api/heartbeats?)~~

Reference:
https://wakatime.com/developers#heartbeats


muety:
Thanks a lot! I'm quite busy at the moment, but will review as soon as I can.

muety:
Great work, thanks for going about this one!

Date parsing looks fine to me and I don't think we need additional routes either. Just a few remarks:

* [ ] Could you please make the returned data schema analogous to what WakaTime returns (see https://wakatime.com/developers#heartbeats, https://wakatime.com/api/v1/users/current/heartbeats?date=2022-01-30)? `{ "data": [...], "start": ..., ... }`? Omit fields that Wakapi doesn't have data for (like `cursorpos`, etc.), but at least make the returned JSON approximately the same structure. You can find inspiration at https://github.com/muety/wakapi/tree/master/models/compat/wakatime/v1.
* [ ] Please move the route definition from `routes/api/heartbeat.go` to `routes/compat/wakatime/v1/heartbeat.go`, where all other "compat"-endpoints reside. Only `POST /heartbeat` is a bit special in that the Wakapi- and the WakaTime endpoint accept exactly the same data format, but for all others, let's keep them in the `compat` directory.
* [ ] Could you please also add a very simple API test case to the Postman testing collection? It's a bit cumbersome, as you will have to install Postman, import the [collection](https://github.com/muety/wakapi/blob/master/testing/Wakapi%20API%20Tests.postman_collection.json), add a test case and save it again. Also, these test cases are not independent of each other, i.e. their execution order matters. I'd still appreciate to have a very simple integration test case for the new endpoint, though. Also see https://github.com/muety/wakapi#api-tests.

Again, thanks a lot for your efforts!

YC:
I somehow missed the example output...
Will fix!

repositories/heartbeart.go
```diff
@@ -77,6 +78,19 @@ func (r *HeartbeatRepository) GetAllWithin(from, to time.Time, user *models.User
 	return heartbeats, nil
 }
 
+func (r *HeartbeatRepository) GetAllWithinUTC(from, to time.Time, user *models.User) ([]*models.Heartbeat, error) {
```
muety, line 81: A few thoughts about this. First, what was your intention with this method? Are you sure this actually does, what you want? In my understanding, this might lead to errors. At least in the case of MySQL, the database connection has a timezone set, which is by default the host's timezone. That is, all dates will be interpreted as that timezone. If you now convert to UTC, won't there be a wrong offset. I once did a bit of research on that topic and took a couple of notes [here](https://github.com/muety/wakapi/blob/7159df30c25a882434185395c7f4927bad34f110/config/db.go#L12), maybe that helps. Would be nice if you quickly shared your thoughts on this timezone conversion.

In any case, I'd rather like the method to be as generic as possible, i.e. `GetAllWithinTZ()` and then pass a timezone as another parameter. Just more flexible.

routes/compat/wakatime/v1/heartbeat.go
```diff
@@ -54,15 +62,23 @@ func (h *HeartbeatHandler) Get(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		w.Write([]byte(conf.ErrInternalServerError))
 		conf.Log().Request(r).Error("failed to retrieve heartbeats - %v", err)
 		return
 	}
-	utils.RespondJSON(w, r, http.StatusOK, heartbeats)
+
+	res := HeartbeatsResult{
+		Data:     heartbeats,
```
muety, line 79: One issue regarding data compatibility again. Of course, the endpoints won't be 100 % compatible. But one thing, that people might find really annoying when coding against Wakapi and WakaTime API at the same time is the different format of `time`. WakaTime returns a timestamp like `1643562171.033` (1), while this endpoint returns `2022-01-21T00:38:46+01:00` (2), for instance. `models.CustomTime` is, admittedly, a bit weird in that is unmarshals from (1), but marshals to (2). One solution would be to simply make `time` a `float` and then map to the time's timestamp for every heartbeat before returning it. Or you might also come up with a better way.

Maybe you can also make use of [`v1.HeartbeatEntry`](https://github.com/muety/wakapi/blob/master/models/compat/wakatime/v1/heartbeat.go) somehow, so we don't have two separate compat heartbeat view models?

muety: Please apologize all these change requests.

repositories/heartbeart.go
YC: According to wakatime API docs, `date` is "Requested day; Heartbeats will be returned from 12am until 11:59pm in user's timezone for this day." 
The intention is to take the date, and query the database for heartbeats between the start and end of the specified date in the user's timezone (from user settings). 
I added this function because using the other function doesn't seem to produce the correct results. 

Some preliminary investigation:
GetAllWithin: `time >= "2022-01-02 00:00:00" AND time < "2022-01-03 00:00:00"`
GetAllWithinUTC: `time >= "2022-01-01 13:00:00" AND time < "2022-01-02 13:00:00"`

Running `sqlite3` manually, 
```
select * from heartbeats WHERE time >= "2022-01-01 13:00:00" AND time < "2022-01-02 13:00:00";
first entry timestamps are 2022-01-01 23:10:38.963+00:00, 2022-01-01 23:10:19.221+00:00
last entry timestamps are 2022-01-02 12:01:42.779+00:00, 2022-01-02 12:01:23.354+00:00

select * from heartbeats WHERE time >= "2022-01-02 00:00:00" AND time < "2022-01-03 00:00:00";
first entry timestamps are 2022-01-02 00:06:05.196+00:00, 2022-01-02 00:05:45.332+00:00
last entry timestamps are 2022-01-02 23:59:43.287+00:00, 2022-01-02 23:59:23.73+00:00
```
When you consider that my timezone is +11, the first appears to be correct. 
However, I do understand the concern around this. 
I'll take a look at your notes and investigate further.

YC:
```go
package main;
import (
    "time"
    "fmt"
)
func main() {
    // User location
    tz, _ := time.LoadLocation("Australia/Melbourne")
    // The input date
    in, _ := time.Parse("2006-01-02", "2022-01-01")

    // utils.StartOfDay(in.In(tz))
    l := in.In(tz)
    d := time.Date(l.Year(), l.Month(), l.Day(), 0, 0, 0, 0, l.Location());

    // GetAllWithinUTC (old), GetAllWithinUnix (new implementation)
    fmt.Println(d.UTC())
    fmt.Println(d.Unix())
    fmt.Println(d.UTC().Unix() == d.Unix())

    // GetAllWithin
    fmt.Println(d.Local())
}
```
Output
```
2021-12-31 13:00:00 +0000 UTC
1640955600
true

2022-01-01 00:00:00 +1100 AEDT
```

An example to show the difference between the `GetAllWithin` and the new implementation.

routes/compat/wakatime/v1/heartbeat.go
YC: > then map to the time's timestamp for every heartbeat before returning it

Can you please elaborate on this?

I've had a look at a sample output from the wakatime api. 
Besides the `time` field, I think a large issue is the various guids, which is non-trivial to replicate (requiring some large migrations, I assume?). However, if we don't use guids, then information will be lost if `v1.HeartbeatEntry` is used. 
So I'd be inclined to change `model.CustomTime` `date` to `float`, if you have no concerns with that?

YC: I've experimented more with this, a couple of possible solutions:
1. Duplicate/change `v1.HeartbeatEntry`, to add additional fields so that information isn't lost. 
2. Duplicate `models.Heartbeat`, change type of `Time` to another `CustomTimeFloat` which deserialize to float and add conversion function to map to this new type which will only be used for this endpoint. 
3. Change `Time` of `models.Heartbeat` to another `CustomTimeFloat`. Downside: Had a taste of this in converting to float, is more reference changes. 
4. Redefine CustomTime to be struct `{ Time: time.Time; UnmarshallToFloat64 bool }`, loop over heartbeats to set that boolean, change Unmarshal function. Downside: I've tried this. It's even more reference changes. 

Now that I understand this issue a bit more, seems like (1) and (2) are what you suggested, will try.

repositories/heartbeart.go
muety: `DATETIME(..., 'unixepoch')` doesn't seem to be a valid MySQL function.

```
2022-02-05T15:06:15.511502714+01:00 [ERROR] failed to retrieve heartbeats - Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(?, 'unixepoch') AND time < datetime(?, 'unixepoch') ORDER BY time asc' at line 1
```

---

Normally, we shouldn't even have to care about converting timezones back and forth, that's something that the driver should handle for you.

---

I'm pretty sure you can just use `GetAllWithin()` as is. Here's an example.
* Using SQLite for now
* I'm in `Euopre/Berlin` (UTC+1)
* E.g. `2022-01-05 23:59` in my time zone corresponds to `2022-01-05 22:59` in UTC
* I have only one heartbeat in my database at exactly that time: 
image
* I modified your endpoint to use `GetAllWithin()` instead of `GetAllWithinUnix()`
* I call `/api/compat/wakatime/v1/users/current/heartbeats?date=2022-02-05`
* According to WakaTime docs, this request shall return all heartbeats that were created on Feb 5th  from my point of view, i.e. in my timezone. Accordingly, I would expect the above heartbeat to be returned. 
* And it actually is:
image
* The server log reads: ```SELECT * FROM `heartbeats` WHERE `heartbeats`.`user_id` = "n1try" AND time >= "2022-02-05 00:00:00" AND time < "2022-02-06 00:00:00" ORDER BY time asc```
* All fine, in my understanding

---

I think the key to this discussion is, that `time` column is of type `TIMESTAMP`, not `DATETIME`. Timestamp type doesn't store any time zone information, it's neither assumed to be UTC nor anything else. It's just the raw date string and is interpreted as whatever the application wants it to interpret as. My Wakapi server, running in `Europe/Berlin` time zone, retrieves `2022-02-05 23:59:00.000` from the database and interprets it as `2022-02-05 23:59:00.000+01:00`. If I sent you my database, everything would be entirely messed up, because although the data doesn't change at all, your server would interpret that date as `2022-02-05 23:59:00.000+11:00` (which is `2022-02-05 12:59:00.000+00:00`). At least in the case of MySQL, I would assume SQLite and Postgres to work similarly.

Analogously for queries. Since the connection's timezone is, in my case, implicitly set to `Europe/Berlin`, a query `... WHERE time >= 2022-02-05 00:00:00 ...` is implicitly actually `... WHERE time >= 2022-02-05 00:00:00+01:00 ...` then, so to say, although this "conversion" never explicitly happens.

Although I'm quite confident with this, I might still be wrong. But then, please prep a very simple example SQLite database that I can use to prove so. Try to avoid issuing raw queries, but test everything "end to end", i.e. insert and query through Wakapi API.

routes/compat/wakatime/v1/heartbeat.go
muety: What about using `v1.HeartbeatEntry` but changing `Time` field to `models.CustomTimeFloat` (or so), which is the same as `models.CustomTime`, but besides unmarshalling _from_ float also marshals _to_ float? The missing fields (e.g. `editor`, etc.) are, strictly speaking, not in WakaTime's view model, so shouldn't be included in ours either. But if you still want to include them I wouldn't mind of you just extended the `v1.HeartbeatEntry` by those few fields.

repositories/heartbeart.go
YC: > DATETIME(..., 'unixepoch') doesn't seem to be a valid MySQL function.

Thanks for the catch, I didn't consider database differences. 

> I'm pretty sure you can just use GetAllWithin() as is. 

Yes, you're right. 
I reverted the change and modified the postman tests to hardcoded dates to specifically test for this and it worked end to end. 

> If I sent you my database, everything would be entirely messed up. 

I will probably need to do more experimentation to fully understand this. 
However, I feel that this behavior is not ideal. 
Perhaps we can create an issue for further investigation/fixes?

muety:
Looks good now! The failing API test is due to the fact that test cases are not dependent on each other and as you create new heartbeats  for your testing, they will be unexpectedly present in later tests as well. Solution would be to either adapt those tests or change the execution order. 

Feel free to merge yourself once all tests are green. 

Increment the `version.txt` to `2.2.0`, but don't yet create a new tag, I'd like to wait for #319 before the next release.

YC:
Thanks, I'll finish resolving your other comment first. 
The `Get all time /all_time_since_today` time test case is a bit weird, since I added entries for time at start of year. 
Is it somehow using created_at instead?

muety:
I haven't looked deeper into it, but I would assume the `WakaTime / Get all time` test is failing because you added more heartbeats, so the total time increases.

muety:
> Feel free to merge yourself once all tests are green.

Actually, let me have another look before merging.

routes/compat/wakatime/v1/heartbeat.go
YC: New example response:
```json
{
  "data": [
    {
      "id": "1",
      "branch": "",
      "category": "",
      "entity": "/home/user1/dev/project1/main.go",
      "is_write": true,
      "language": "Go",
      "project": "wakapi",
      "time": 1640995199,
      "type": "file",
      "user_id": "testuser",
      "machine_name_id": "",
      "user_agent_id": "PostmanRuntime/7.29.0"
    },
    {
      "id": "2",
      "branch": "",
      "category": "",
      "entity": "/home/user1/dev/project1/main.go",
      "is_write": true,
      "language": "Go",
      "project": "wakapi",
      "time": 1641077999,
      "type": "file",
      "user_id": "testuser",
      "machine_name_id": "",
      "user_agent_id": "PostmanRuntime/7.29.0"
    }
  ],
  "end": "2022-01-01T23:00:00Z",
  "start": "2021-12-31T23:00:00Z",
  "timezone": "Europe/Berlin"
}
```

YC:
> I haven't looked deeper into it, but I would assume the `WakaTime / Get all time` test is failing because you added more heartbeats, so the total time increases.

Yes, I was confused by the name of this route. 

Now ready for review again. 
Please squash merge if you're happy.

muety:
Great work! Thanks again for your efforts and your patience.

PR #319, opened 2022-02-08
Remove "Create Account" button when AllowSignup is set to false

### LINKED ISSUE #318: Hide/Disable Sign Up button when registration is disabled for instance
When instance is configured with `security.allow_signup / WAKAPI_ALLOW_SIGNUP = false`, the 'Sign Up' button should be disabled with tooltip or hidden from login page.

notarock:
Entirely remove the "Sign Up" and "Create Account" buttons from "Login" and "Sign up" pages when `security.allow_signup: false` 

`allow_signup: true` :

image
image

`allow_signup:  false` : 

image
image

As I was opening this pull-request, I realized it might have been better to simply disable the `/signup` page entirely when `allow_signup` is set to false. Also, it could be better to disable the "sign up" button (Without removing it) and let the user know the administrator has disabled signing up on that instance (Via a tooptip maybe?)

Let me know if you think we should redirect `/signup` when it is not allowed. Maybe we should use HTTP code `404` or `403` ?

Closes #318


muety:
Thanks for this contribution! 

> Also, it could be better to disable the "sign up" button (Without removing it)

Yes, I'd actually prefer that option. Disable the button and add a quick tooltip (using HTML `title` attribute), please.

> Let me know if you think we should redirect /signup when it is not allowed.

I don't think that's necessary. If a user purposefully navigates to `/signup`, they shouldn't be surprised when seeing an error message after submitting the form.

notarock:
> Any news on this?

Sorry, life got busy for a second. I'll do the proposed fix very soon

notarock:
I added a new CSS class, `btn-disabled` , that look like `btn-primary` but is slighly darker. After the changes, it looks like this:

### Signup true

image 
image

### Signup false

image
image

image

notarock:
A bit unrelated, but I think it would be a nice touch to add a [editorconfig file](https://editorconfig.org/) to this repository. Some text editors add a new line at the end of file, others don't... Editorconfig enforces some basic rules for consistency, and that would prevent situations like this PR where simply saving a file with a different text editor causes unrelated changes to appear.

muety:
Awesome, great work!

> I think it would be a nice touch to add a editorconfig file to this repository

I agree. Feel free to make a proposal, once you get to it! For the Go part, there shouldn't be much ambiguity regrding code format due to using `go fmt`, but for the JS and HTML parts it might be beneficial.

YC:
> that would prevent situations like this PR where simply saving a file with a different text editor causes unrelated changes to appear

Perhaps a `gitattributes` file with `eol` directives would be good for this.

YC:
Added in #320

PR #320, opened 2022-02-17
ref: add gitattributes file, remove unnecessary unicode characters

YC:


views/head.tpl.html
```diff
@@ -1,5 +1,5 @@
 <head>
-    <title>Wakapi – Coding Statistics</title>
+    <title>Wakapi - Coding Statistics</title>
```
muety, line 2: These "long" dashes were actually intended. Is there a good reason to replace them?

testing/run_api_tests.sh
```diff
@@ -30,7 +30,7 @@ done
 echo ""
 
 echo "Running test collection ..."
-newman run "Wakapi API Tests.postman_collection.json"
+newman run "wakapi_api_tests.postman_collection.json"
```
muety, line 33: Won't this break the tests, as the file is actually named like so?

views/head.tpl.html
YC: I reverted this file, but happy to revert the `.go` error messages also if you want. 
An argument might be that it's more standard, but it's a rather weak reason.

PR #326, opened 2022-03-02
fix: href to a 404 when service on a subpath

daief:
click project detail will redirect to a not exist page, when the service runs with a base path.

For example, the base path is `wakatime`,  and the dashboard uri will be `/wakatime/summary`. When click project detail, page will be redirect to `/wakatime/wakatime/summary?project=demo` but the correct detail page is `/wakatime/summary?project=demo`.

And i think `pushing a history stack` is better than `replace`, so that can back to dashboard by backwards.


muety:
Thanks for bringing this up!

`a.href` is not existing. Instead I was assuming you meant `url.href`. Changed it accordingly, please check.

PR #332, opened 2022-03-05
typo fix

Zxilly:


YC (approved):

muety (commented):
That is actually intended, plural form of OS.

PR #347, opened 2022-03-27
feat: allow toggling vibrant color for all charts

### LINKED ISSUE #343: Option to switch on vibrant colors for editors and OS'
Hi @muety, are there any plans to bring back colorful editor and os charts? any particular reason for changing these into generic green gradients?

ref: https://github.com/muety/wakapi/commit/44a2e609fb83414d4532182772d4ce3f2fb6d10e

bdeshi:
- supports new config key `app.vibrant_color` or env `WAKAPI_VIBRANT_COLOR`
- updates and extends `data/colors.json` with editor and os colors
- fixes #343


bdeshi:
hi @muety, please review

muety:
Great contribution, thanks a lot! To be honest, I was thinking to have this vibrant color toggling more on a per-user basis, rather than globally for the whole server. Like an option somewhere on the settings or a small, subtle toggle switch on the dashboard page or inside a chart container to swap between "muted" and "vibrant" colors. That switch would simply set a boolean value in the user's browser's localStorage (would not actually persist that server-sided), depending on which the respective color palette is chosen. 

Would you be willing to change your current implementation accordingly? Large parts can still be kept of course.

bdeshi:
> Great contribution, thanks a lot! To be honest, I was thinking to have this vibrant color toggling more on a per-user basis, rather than globally for the whole server.

ok sure, that sound better. i'll add follow-up commits here.

> That switch would simply set a boolean value in the user's browser's localStorage (would not actually persist that server-sided)

i guess this pref can be added to the users table. i won't want to toggle this setting every time i login to wakapi from a new browser.

muety:
> i guess this pref can be added to the users table. i won't want to toggle this setting every time i login to wakapi from a new browser.

Since it would ideally be literally just one button click, I think it's a bit of overkill to persist that setting to the database. Do whatever you prefer, though.

muety:
Any news here? :)

bdeshi:
> Any news here? :)

hi sorry got busy with other stuff, will update this later if it's okay.

bdeshi:
@muety hi how's this now?

muety:
Thanks a lot! This is exactly how I imagined it! Only did very few code changes in addition.

PR #364, opened 2022-04-24
chore: remove hard-coded volume from Dockerfile

bdeshi:
the Dockerfile has a `VOLUME /data` defined. However, there are multiple reasons to skip this:

1) volumes can be (are expected to be?) defined when running with `docker run` or `docker-compose`
2) this volume is unnecessary and not useful if sqlite db backend is not being used
3) it is easier to define a volume as needed than to _undefine_ a hard-coded volume
4) this creates lots of stale volumes during repeated builds with docker or compose

also, [the readme](https://github.com/muety/wakapi#-option-3-use-docker) already shows a `docker run` command complete with volume definition, and the `docker-compose.yml` in this PR defines a persistent volume for db data.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,8 +54,6 @@ ENV WAKAPI_ALLOW_SIGNUP 'true'
 
 COPY --from=build-env /app .
 
-VOLUME /data
-
 EXPOSE 3000
 
 ENTRYPOINT /app/entrypoint.sh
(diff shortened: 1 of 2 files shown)


YC:
Not familiar with volumes. 

From reading https://stackoverflow.com/questions/39317669/dockerfile-persist-data-with-volume, it seems that the `VOLUME` instruction in `Dockerfile` does not result in a persistent volume. Instead, a fresh volume is created for each instance of `docker run`. In this case, I agree that `VOLUME` should be removed from `Dockerfile`. 

I also agree that the db should have a volume specified in `docker-compose.yml`. 
One minor comment in that `{}` may not be necessary on line 29, but the changes look good to me. 

Thanks @bdeshi!

muety:
Looks good to me. Thank you!

Issue #351, opened 2022-04-08
Incorrect Instructions on first login to wakapi.dev (possible solution)


justEstif:
I thought I should inform you that the instructions in the wakapi.dev on first login were wrong. I had followed its instructions, but it wasn't receiving any heartbeats. 
The problem was that it said set: 
api_url = https://wakapi.dev/api

It should say
api_url = https://wakapi.dev/api/heartbeat

muety:
Both should work. If you add `/heartbeat` it will simply be cut off by the CLI again, see [here](https://github.com/wakatime/wakatime-cli/blob/0d5c0ad3a26ae4bb903f3d5fbdbd2a21ecfef980/cmd/params/params.go#L177).

I'm personally using `https://wakapi.dev/api` in my `.wakatime.cfg` and it works.

justEstif:
Thanks for the quick reply. I tested it out the original instructions but it didn't not work. The heartbeat count didn't go up, nor did anything else. Once I went back to my suggestion, it appears to work. I can't really offer any more reasoning as I am ill equipped, but thought I should share it here.

muety:
Can you try to set it back to `/api`, do a bit of coding, and then explicitly invoke the CLI?

```bash
~/.wakatime/wakatime-cli --verbose --sync-offline-activity 1000
```

Also, are there any related log messages in `~/.wakatime.log`?

justEstif:
I didn't find anything in the logs that suggested something recent. I think I might have changed some settings when I was setting up wakatime for vim and/or emacs. Thanks for the help any way.

PR #365, opened 2022-04-26
Fix wrong url in setup instructions

brorbw:
I've spent hours trying to figure out why I got a 401 and 404. The issues was that the help test on the index page is not the same as the one in the readme

### DIFF
--- a/views/summary.tpl.html
+++ b/views/summary.tpl.html
@@ -198,7 +198,7 @@
                     # <strong>Step 2:</strong> Set your ~/.wakatime.cfg to this:<br><br>
                     <!-- https://github.com/muety/wakapi/issues/224#issuecomment-890855563 -->
                     [settings]<br>
-                    api_url = <span class="with-url-inner">%s/api</span><br>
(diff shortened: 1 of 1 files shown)


muety:
Are you sure this was the issue? Something similar was already discussed in https://github.com/muety/wakapi/issues/351, and both `/api` and `/api/heartbeat` should work equally. I'm using only `/api` myself.

brorbw:
Quite sure. Changing the `api_url` in the config to `api/heartbeat` fixed the issue. Looking at the logs the server responds with `201` and `200` for request on `api/heartbeat` where it responded with `404` and `401` for the `api` endpoint.

brorbw:
Too add to that. Inconsistencies smells a bit.

muety:
Can you please try to do [this](https://github.com/muety/wakapi/issues/351#issuecomment-1093016016) and see if you get 404's then? 

Also, which version of `wakatime-cli` are you running?

muety:
Good to hear. Yeah, the legacy CLI was stricter with regard to API URL paths.

PR #372, opened 2022-05-11
add Mayhem for API as a github workflow

dlowe:
Add a github action to run [Mayhem for API](https://mayhem4api.forallsecure.com/) (which is a free automated testing tool for http APIs) against the api: given the swagger spec and api server, it generates and runs random test payloads, looking for exceptional behavior.

Disclosure: working on this tool is my day job!

As with any github workflow, test results can be consumed through github. Alternatively, Mayhem for API also has a dashboard, over here: https://mayhem4api.forallsecure.com/mayhemheroes/wakapi.

Currently this finds a bunch of Internal Server Errors that can be pretty easily triggered, an Authentication Bypass (unauthenticated success against an endpoint which the spec says requires auth), and a (seemingly consistently reproducible) timeout.

To make it work in the upstream repo, there's a bit of out-of-PR effort to integrate:

- sign up for a free mapi account, creating an muety organization
- create a mapi service account within the muety organization
- modify the mapi.yaml action in this PR to point at your mapi organization
- add the service account's API token to github as a repository secret named MAPI_TOKEN

### DIFF
--- a/.github/workflows/mapi.yml
+++ b/.github/workflows/mapi.yml
@@ -1,0 +1,52 @@
+name: 'Mayhem for API'
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.18
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: go get
+
+    - name: Build
+      run: GO111MODULE=on go build -v .
+
+    - name: start wakapi
+      run: ./wakapi --config config.default.yml &
+
+    - name: create a trivial testing user
+      run: sqlite3 wakapi_db.db "insert into users (id, api_key) values ('mapi', 'test-api-key')"
+
+    - name: Run Mayhem for API
+      uses: ForAllSecure/mapi-action@v1
+      continue-on-error: true
+      with:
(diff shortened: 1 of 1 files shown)


muety:
Hi @dlowe, thanks a lot for this contribution! Let me have a few comments on it.

The dashboard link you posted seems to be broken. Would be cool to get a glance of what Mayhem has found in Wakapi, how the dashboard looks like, etc. before integrating it.

Generally, I'm a bit reluctant to integrate third-party tools. Especially if they're not open-source and especially if I didn't explicitly choose them. Mayhem actually looks pretty cool at first sight, though! Could you maybe leave a few words about why to use Mayhem over other tools (and what other tools there are in the first place)? I'd love to have a bit of a "comparison".

Also, what does the _Upload SARIF file_ action do?

muety:
Thanks a lot for this very detailed explanation and comparison! I think I quite like Mapi. What I didn't fully get yet is the SARIF uploading. Where is that file uploaded to? To Mapi's API or to GitHub? What I wouldn't really want to have is things like the tool automatically creating GitHub tickets or it committing any kind of files to the repo or so.

dlowe:
> Thanks a lot for this very detailed explanation and comparison! I think I quite like Mapi. What I didn't fully get yet is the SARIF uploading. Where is that file uploaded to? To Mapi's API or to GitHub? What I wouldn't really want to have is things like the tool automatically creating GitHub tickets or it committing any kind of files to the repo or so.

Oh, I see. No, nothing like that. Mapi saves the SARIF report to the disk of the container in which the github action is running, and then the codeql step picks it up. It might be possible to configure codeql to turn SARIF into github issues, but that's certainly not the default.

muety:
Again, thanks a lot for your contribution!

PR #377, opened 2022-05-28
elaborate on cloud server API URL

NChechulin:
Personally, for me, it was slightly unclear which URL has to be pasted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ Wakapi relies on the open-source [WakaTime](https://github.com/wakatime/wakatime
 ```ini
 [settings]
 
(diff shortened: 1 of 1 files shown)


muety:
Actually, `/api/heartbeat` was for the old version of wakatime-cli. It still works for the new version for compatibility reasons, but `/api` is preferred. You can please quickly change both references accordingly? Thanks!

NChechulin:
@muety Sure I can!
Should the `http://localhost:3000/api/heartbeat` be changed to `http://localhost:3000/api` as well?

muety:
Yes.

Issue #355, opened 2022-04-13
API Tests on Linux randomly failing on CI


muety:
When I pushed a change last night, the Linux build was failing due to an assertion error in the API tests. I retried twice, but no luck. This morning the tests went fine. I'd assume some time-zone related issue. All failed attempts were between 0 and 1 am CET time.

These are the four attempts, last one is this morning's, which was successful eventually:
* https://github.com/muety/wakapi/actions/runs/2157817036/attempts/1
* https://github.com/muety/wakapi/actions/runs/2157817036/attempts/2
* https://github.com/muety/wakapi/actions/runs/2157817036/attempts/3
* https://github.com/muety/wakapi/actions/runs/2157817036/attempts/4

I'd be happy if somebody is keen to look into this. Not high prio, though.

PR #382, opened 2022-06-28
fix: admin users can't fetch other user data

Enthys:


muety:
Thanks for your contribution. Just out of interest, can you please give some background / possible use cases, why you would want to have this? Of course, if an admin user is also the server admin at the same time, she'll have unlimited access to all data anyway. But there might be admin users, who don't have database access, so why would you want to additionally empower them to read and write data on behalf of another user with the API?

Enthys:
The change is primarily for the `user` routes, because the `user` route argument is very redundant given that not even an admin can access other user's data. 
The reason, at least for us, for providing this access to user information to the admin user is because we need a way to extract the developer's time, how much is worked on a project, on which branch, etc. and currently there is no other easy way to do this except for either prompting the developer to provide this information, give account credentials to do it ourselves or find a way to extract this information from the database(in this last case the admin user could not even be familiar with how to work with databases)

muety:
I see, thanks for providing clarity on this, makes much more sense to me now. So the admin having access to individual users' data especially comes into play when using Wakapi as a time-tracking system in teams. 

Given that an admin anyway has ability to create, delete, etc. users, accessing the heartbeats API on their behalf is reasonable, I think.

muety:
To accomplish what you want, I think you'll need to return `requestedUser` in line 45, though. 

If you don't mind spending even a little more time on this, I'd really like to have basic unit tests for this method to verify it behaves correctly. You can use [`net/http/httptest`](https://pkg.go.dev/net/http/httptest?utm_source=godoc#example-ResponseRecorder) and the `UserServiceMock` for this. I'd appreciate your support on this! But don't feel obligated, otherwise I'll add the tests myself.

muety:
Awesome contribution, thanks a lot! 

API tests are currently failing, but also on `master`, so not related to your changes. It might be relates to https://github.com/muety/wakapi/issues/355. I'll try to re-run them later and will otherwise have to deeper investigate that issue. Will merge this once I got the API tests fixed.

PR #388, opened 2022-07-09
feat: automatic version.txt updates

### LINKED ISSUE #387: Set version from Git tag using hook
I'd assume there is a way to have a Git hook that automatically adapts `version.txt` when a tag is set. Would be helpful to prevent issues like #386.

YC:
Sets `version.txt` automatically on publish
Resolves #387 

Version is derived in order: tag, branch, commit
Command adapted from https://stackoverflow.com/a/55276236


.github/workflows/release.yml
```diff
@@ -33,6 +33,13 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Set version
+      shell: bash
```
YC, line 37:
```
The default shell on non-Windows platforms with a fallback to sh. When specifying a bash shell on Windows, the bash shell included with Git for Windows is used.
```
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

muety:
Hey, just wanted to let you know that I haven't forgot about you! I'm on holidays right now and will review this afterwards.

.github/workflows/release.yml
muety: Very interesting!

PR #397, opened 2022-08-19
Simplify Dockerfile

YC:


Dockerfile
```diff
@@ -1,26 +1,17 @@
-# To build locally: docker buildx build . -t wakapi --load
-
-# Preparation to save some time
-FROM --platform=$BUILDPLATFORM golang:1.18-alpine AS prep-env
```
muety, line 4: Do you know that this `--platform` tag was used for?

Dockerfile
```diff
@@ -41,16 +32,16 @@ WORKDIR /app
...
+ENV ENVIRONMENT=prod \
+    WAKAPI_DB_TYPE=sqlite3 \
+    WAKAPI_DB_USER='' \
+    WAKAPI_DB_PASSWORD='' \
+    WAKAPI_DB_HOST='' \
+    WAKAPI_DB_NAME=/data/wakapi.db \
+    WAKAPI_PASSWORD_SALT='' \
+    WAKAPI_LISTEN_IPV4='0.0.0.0' \
+    WAKAPI_INSECURE_COOKIES='true' \
+    WAKAPI_ALLOW_SIGNUP='true'
```
muety, line 45: I realized the container crashes with a confusing error message when using SQLite and not mounting the volume to `/data`. While using a volume is highly encouraged for data persistence, the container should still be able to without one. Could you please add this line here:

```
RUN mkdir /data
```

muety (changes requested):

Dockerfile
YC: To run the `prep-env` stage only once (on the host platform) when building for multiple targets for release, from #293 . 
However, it seems that `go build` still downloads packages. As this is only used during building releases, I feel that the minor performance enhancement (if any exists) is not worthwhile when compared to readability and ease of building with the regular `docker build`.

Dockerfile
YC: Yes, I came across this also. 
I was thinking that it would be better to verify this in code, as it could be a valid occurrence outside Docker images.

muety: I think outside of the container it's fine, because there the default configuration doesn't result in a crash.

Default config for the SQLite file is `wakapi_db.db` (current directory), which is always available.

But in the container, the default path is `/data/wakapi.db`, which is not available by default and you would have to explicitly change the environment variable to fix this.

Dockerfile
```diff
@@ -1,26 +1,17 @@
...
-RUN apk add --no-cache gcc musl-dev
-
-WORKDIR /src
-COPY --from=prep-env /src .
+ADD ./go.mod ./go.sum ./
+RUN go mod download
+ADD . .
 
-RUN go build -v -o wakapi
+RUN go build -ldflags "-s -w" -v -o wakapi main.go
```
muety, line 14: Didn't know about these btw., very cool :+1:.

YC:
After the latest commit
```
$ docker run -it wakapi:latest
2022-08-19T11:49:42.562048777Z [WARN ] with sqlite, only a single connection is supported
2022-08-19T11:49:42.562286201Z [ERROR] unable to open database file: no such file or directory
2022-08-19T11:49:42.562296132Z [ERROR] could not open database
panic: could not open database

goroutine 1 [running]:
github.com/emvi/logbuch.(*StandardFormatter).Pnc(0xc000188000?, {0xcb60bd?, 0xcb60bd?}, {0x0?, 0x0?, 0x0?})
	/go/pkg/mod/github.com/emvi/logbuch@v1.2.0/standard_formatter.go:57 +0x3f
github.com/emvi/logbuch.(*Logger).Fatal(0xc000188000, {0xcb60bd, 0x17}, {0x0, 0x0, 0x0})
	/go/pkg/mod/github.com/emvi/logbuch@v1.2.0/logger.go:132 +0x77
github.com/emvi/logbuch.Fatal(...)
	/go/pkg/mod/github.com/emvi/logbuch@v1.2.0/funcs.go:55
main.main()
	/src/main.go:126 +0x379
```
mkdir upcoming...

YC:
image

PR #402, opened 2022-09-06
Add more architectures to release binaries

### LINKED ISSUE #401: Provide arm64 releases
Currently, we're building x64 binaries for Linux, Windows and Mac. Especially now, that ARM-based machines gain in popularity (Mac M1, AWS Graviton2, ...) we should also provide builds for arm64 architecture. This is also a prerequisite for #400.

Easiest would probably be to add an `arch` dimension to our build matrix, see https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context. 

Alternatively, we could consider switching to https://github.com/goreleaser/goreleaser-action (inspired by https://github.com/screego/server/blob/master/.github/workflows/build.yml), but I don't know if that's worth the effort. 

@YC Would you mind taking a look into this, as you're way more familiar with GHA than I am?

YC:
Successful execution of first commit: https://github.com/muety/wakapi/actions/runs/2999385641
Not super confident about the release action, but hopefully works out...

Resolves #401

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,27 +1,20 @@
 name: Release
 
 on:
-  release:
-    types:
-      - published
+  push:
+  # release:
+  #   types:
+  #     - published
 
 jobs:
   release:
     name: 'Build, package and release to GitHub'
-
+    runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


muety:
Thanks a lot! Looks good to me :+1: I guess we'll only see this in action with the next release then? :thinking:

PR #404, opened 2022-09-07
Fix release by enabling cgo

### LINKED ISSUE #403: SQLite stopped working
**Describe the bug**
I upgraded from a previous  version (I don't remember which one exactly) and SQLite doesn't work anymore.

```
2022-09-07T00:25:24.736313+01:00 [WARN ] with sqlite, only a single connection is supported
2022-09-07T00:25:24.7368331+01:00 [ERROR] Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work. This is a stub
2022-09-07T00:25:24.7368331+01:00 [ERROR] could not open database
panic: could not open database

goroutine 1 [running]:
github.com/emvi/logbuch.(*StandardFormatter).Pnc(0xc0000da000?, {0xde9718?, 0xde9718?}, {0x0?, 0x0?, 0x0?})
        /home/runner/go/pkg/mod/github.com/emvi/logbuch@v1.2.0/standard_formatter.go:57 +0x3f
github.com/emvi/logbuch.(*Logger).Fatal(0xc0000da000, {0xde9718, 0x17}, {0x0, 0x0, 0x0})
        /home/runner/go/pkg/mod/github.com/emvi/logbuch@v1.2.0/logger.go:132 +0x77
github.com/emvi/logbuch.Fatal(...)
        /home/runner/go/pkg/mod/github.com/emvi/logbuch@v1.2.0/funcs.go:55
main.main()
        /home/runner/work/wakapi/wakapi/main.go:126 +0x379
```

**System information**
* Wakapi version 2.3.8 (amd64)
* Operating system Windows 11
* Database (SQLite, MySQL, ... ?) SQLite

YC:
Resolves #403, #405.

The build is fine, but it's unclear whether the cross compilation works correctly without further testing.
https://github.com/muety/wakapi/actions/runs/3007275529

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,8 +64,7 @@ jobs:
       shell: bash
       run: |
         GOOS=${{ matrix.GOOS }} GOARCH=${{ matrix.GOARCH }} \
-          CC=${{ matrix.CC }} CXX=${{ matrix.CXX }} CGO_ENABLED=1 \
-          go build -v ../
+          CC=${{ matrix.CC }} CGO_ENABLED=1 go build -v ../
 
     - name: Compress working folder (Windows PowerShell)
       working-directory: ./dist
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
       shell: bash
       run: |
         GOOS=${{ matrix.GOOS }} GOARCH=${{ matrix.GOARCH }} \
-          CC=${{ matrix.CC }} CGO_ENABLED=1 go build -v ../
+          CC=${{ matrix.CC }} CGO_ENABLED=1 go build -v -ldflags '-w -s' ../
 
     - name: Compress working folder (Windows PowerShell)
       working-directory: ./dist
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,19 +1,37 @@
 name: Release
 
 on:
-  release:
-    types:
-      - published
+  push:
+  # release:
+  #   types:
+  #     - published
 
 jobs:
   release:
     name: 'Build, package and release to GitHub'
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        GOOS: [darwin, windows, linux]
-        GOARCH: [amd64, arm64]
+        include:
+          - platform: ubuntu-latest
+            GOOS: linux
+            GOARCH: amd64
+            CC: gcc
+          - platform: ubuntu-latest
+            package: gcc-aarch64-linux-gnu
+            GOOS: linux
+            GOARCH: arm64
+            CC: aarch64-linux-gnu-gcc
+          - platform: windows-latest
+            GOOS: windows
+            GOARCH: amd64
+          - platform: macos-latest
+            GOOS: darwin
+            GOARCH: amd64
+          - platform: macos-latest
+            GOOS: darwin
+            GOARCH: arm64
+    runs-on: ${{ matrix.platform }}
 
     steps:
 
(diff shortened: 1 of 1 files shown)


YC:
> Stupid question, but is there a way to download build artifacts from a GH Actions run?

It's possible to use `actions/upload-artifact@v3` (as opposed to releasing it). 
Should have thought of that 🙂

muety:
Would you mind including that, please? :) Again, no rush!

And for my understanding: I was wondering why the _Upload built executable to Release_ task is not run in the above run. Is it, because that action "internally" limits itself to when triggered by new releases?

YC:
> Would you mind including that, please? :) Again, no rush!
> 
> And for my understanding: I was wondering why the _Upload built executable to Release_ task is not run in the above run. Is it, because that action "internally" limits itself to when triggered by new releases?

Hmm I actually don't know what happens when you try to upload release files outside of a release, but I deleted it along with enabling build on push since I needed to test it. These changes are reverted by the latest commit.

YC:
See https://github.com/muety/wakapi/actions/runs/3007539130
Please pardon the double zipping from the `upload-artifact` action.

muety:
> Hmm I actually don't know what happens when you try to upload release files outside of a release

From my understanding, the `upload-artifact` is entirely decoupled from releases. It simply persists the artifacts on GitHub infrastructure so they are available for download after the job has finished. I think you could always upload artifacts (mainly for CI debugging purposes) and, in case of a new release, additionally add them to the release. I might be wrong. Double check and potentially add the uploads action if you're keen. Otherwise this is not of high priority, though.

---

I checked `linux_amd64` and `windows_amd64` and they're both running fine. Did you check the Mac version? If that one works as well, I'd probably suggest to (1) merge this and (2) replace the current release assets with the ones above (only unzipped by one level).

muety:
Also, we could potentially add these these additional compilter flags, which result in smaller binaries, see https://github.com/muety/wakapi/blob/master/Dockerfile#L14.

YC:
> > Hmm I actually don't know what happens when you try to upload release files outside of a release
> 
> From my understanding, the `upload-artifact` is entirely decoupled from releases. It simply persists the artifacts on GitHub infrastructure so they are available for download after the job has finished. I think you could always upload artifacts (mainly for CI debugging purposes) and, in case of a new release, additionally add them to the release. I might be wrong. Double check and potentially add the uploads action if you're keen. Otherwise this is not of high priority, though.

Yes, I'm using `upload-artifact` in that linked execution to upload build artifacts, which is on a different branch. 
It is possible to both `upload-artifact` and `softprops/action-gh-release`, but there's probably not much benefit - since to debug it, you'd always need to edit the release file anyway to make it run on `push`. 
Don't think there's a need to worry about the double zip, since we didn't have issues with `action-gh-release`. 

> 
> I checked `linux_amd64` and `windows_amd64` and they're both running fine. Did you check the Mac version? If that one works as well, I'd probably suggest to (1) merge this and (2) replace the current release assets with the ones above (only unzipped by one level).

I've checked both the Linux and Windows, including the arm64. Don't have a Mac. 
I think there should be no issues with `darwin_amd64`, but `darwin_arm64` uncertain. 

> Also, we could potentially add these these additional compilter flags, which result in smaller binaries, see https://github.com/muety/wakapi/blob/master/Dockerfile#L14.

Will investigate later today.

YC:
The addition of `-s -w` didn't seem to have caused any issues. 
https://github.com/muety/wakapi/actions/runs/3013287915

muety:
> I think there should be no issues with darwin_amd64, but darwin_arm64 uncertain.

I had a friend check the MacOS x64 release and works fine :+1: 

> The addition of -s -w didn't seem to have caused any issues.

Binaries are half the size, awesome!

@YC Thank you so much for your efforts with this! 

Do you think there is anything against replacing the current release's files with ones from the build you linked above? Otherwise I will do so.

PR #419, opened 2022-10-07
Tailwind 3 & Footer alignment

YC:


muety:
Great contribution, thank you! Only thing: the green is somewhat "greener" now :thinking:.

Before: `#047857`

After: `#15803D`

Could you please fix that, as the original green has kind of become part of Wakapi's "CI". Feel free to merge yourself afterwards.

PR #425, opened 2022-11-05
Persistent summary time interval

### LINKED ISSUE #391: Add an option to choose which timeframe should be displayed by default
Add an option, either in the web interface (it could be an option different for each user or the admin choose it) or in the environment variable. And allow to  choose which timeframe should be displayed by default.
For exemple always display this week instead of the current day.

Daste745:
These changes allow the back-end code to read a `wakapi_summary_interval` cookie and use it when the `interval` and `from` query params are missing.

Unfortunately, I didn't manage to send back a `Set-Cookie` header to persist the data on client-side, so some help here would be appreciated.

Closes #391.

### DIFF
--- a/models/shared.go
+++ b/models/shared.go
@@ -12,9 +12,10 @@ import (
 )
 
 const (
-	UserKey       = "user"
-	ImprintKey    = "imprint"
-	AuthCookieKey = "wakapi_auth"
+	UserKey               = "user"
+	ImprintKey            = "imprint"
+	AuthCookieKey         = "wakapi_auth"
+	PersistentIntervalKey = "wakapi_summary_interval"
 )
 
 type MigrationFunc func(db *gorm.DB) error
--- a/routes/summary.go
+++ b/routes/summary.go
@@ -4,10 +4,12 @@ import (
 	"github.com/gorilla/mux"
 	conf "github.com/muety/wakapi/config"
 	"github.com/muety/wakapi/middlewares"
+	"github.com/muety/wakapi/models"
 	"github.com/muety/wakapi/models/view"
 	su "github.com/muety/wakapi/routes/utils"
 	"github.com/muety/wakapi/services"
 	"github.com/muety/wakapi/utils"
+	"fmt"
 	"net/http"
 )
 
@@ -49,8 +51,9 @@ func (h *SummaryHandler) GetIndex(w http.ResponseWriter, r *http.Request) {
 			q.Set("interval", "today")
 		}
 		r.URL.RawQuery = q.Encode()
-	} else {
-		// TODO: Add a `Set-Cookie: interval` header to persit it on the front-end
+	} else if q.Get("interval") != "" {
+		headerValue := fmt.Sprintf("%s=%s", models.PersistentIntervalKey, q.Get("interval"))
+		w.Header().Add("Set-Cookie", headerValue)
 	}
 
 	summaryParams, _ := utils.ParseSummaryParams(r)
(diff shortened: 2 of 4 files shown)


muety:
Thanks for this contribution! Looks good so far :+1: For setting a cookie, check the `ResponseWriter`'s [`SetCookie()`](https://pkg.go.dev/net/http#SetCookie) method. Hope that helps!

muety:
Hey @Daste745, any updates on this?

Daste745:
Yes, sorry, I've been focusing on other things lately, so I didn't have much time to work on this PR 😔. I'll finish it as soon as I find some time.

muety:
No worries :+1:

Daste745:
Alright, the response should now contain a `Set-Cookie` header with the used summary. It only sets the header if an explicit summary interval query param was provided.

muety:
While testing this before having looked at the code, I got confused for a second, because the URL in the summary page didn't match what was actually shown on the page anymore. For example, if you select `month` interval, then move to a different page and get back to the summaries page, it will correctly show the month-data, but the URL will be just `/summary`, instead of `/summary?interval=month`. This would result in ambiguity, e.g. when wanting to share the current view by link.

Technically, your implementation works and I highly appreciate your contribution! But I think we can get this even a little cleaner, still. 

You added a bit of logic to check the cookie from inside the client-side JavaScript code and accordingly set the interval. What you could do now - in order to fix the above URL issue - is to also update `window.location.search` from the JavaScript accordingly. But I think there is a cleaner way even.

I propose to actually make this a "backend-only" change, i.e. keep the JavaScript out entirely. The process would look like this: 
1. Server receives request at `/summary`
2. It checks if `interval` URL parameter is present and, if yes, shows the according summary
3. If not, it moves on to checking if that newly introduced cookie is present and, if, yes returns a redirect (see [here](https://github.com/muety/wakapi/blob/master/routes/login.go#L177) for an example) to `/summary?interval=<the-interval-from-the-cookie>`. This would jump the process back to (1), but this time, an `interval` param will be found in (2)
4. If no param and no cookie are present, `"today"` is used as a fallback, just like before

This might be a little less efficient from a technical perspective, because there is a redirect involved. But I think it's a lot cleaner and all previous interval-related logic can stay just as is.

Do you think that makes sense?

Sorry for requesting change here! I'm aware that this is probably a bit picky, especially since your current code works perfectly fine. But I'd still like to include this change. No stress doing so! 

Thank you! :pray:

Daste745:
Hi, thanks for the review!

I aggree on your points here, the current implementation introduces ambiguity, which is not always desirable. I like the redirect approach. We could even apply that to the `/` -> `/summary` redirect and go directly to `/summary?interval=<cookie_interval_if_exists>`. 

Regarding the perfomance, if we redirect early enough, it shouldn't be very noticeable, as an empty response with a redirect should be very small and quick on the client side. If this proves to be false, we could also use the [History.replaceState()](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) browser API, although I don't think this will be needed.


> Sorry for requesting change here! I'm aware that this is probably a bit picky, especially since your current code works perfectly fine. But I'd still like to include this change. No stress doing so!

No stress with that, I'll make the changes. Thanks for the example, it'll make it easier for me!

Daste745:
Hi, I updated the logic to redirect to a correct summary url (with an `interval` query param) when it encounters a `/summary` request without the `interval` param and the `wakapi_summary_interval` cookie is set.

I also removed the (now unnecessary) front-end logic, which read the cookie value, from the time picker and summary page's `getCurrentInterval` function.

Sorry this took so long, I hope it's not an issue. 😅

muety:
Awesome contribution, thanks a lot! :raised_hands:

PR #430, opened 2022-11-16
fix: ignore error "starttls command has been already sent"

xiecang:


muety:
Very cool, thanks :+1:

Issue #407, opened 2022-09-08
Upgrade Testing


YC:
An issue that has occurred in the past is the failure of migrations given different target dbs (sqlite, postgres, mysql). 
Most of these should be detectable by doing some automated testing. 
i.e. Download and launch previous version of wakapi (and somehow exit), run new version of wakapi.

muety:
Thanks for bringing this up. The question is what exactly to test here.

We currently have (a few) unit tests and (a bit more) API tests. But they are only run on Linux and only with SQLite database. I think primary goal would be to extend testing in a way that especially the cross-database compatibility is verified with every CI run. Running the API test suite on MySQL and Postgres as well would probably suffice. 

However, I don't know about good practices to automate testing with a fully-fledged database server. CI would have to:
- Start a temporary MySQL / Postgres server
- Import initial test data / fixtures (from an SQL file) 
- Start Wakapi
- Run the _newman_ tests
- Tear everything down

Has anyone got any experiences about automating this in CI (specifically GH Actions)?

muety:
Just randomly came across https://github.com/dolthub/go-mysql-server. Might actually be interesting, at least for testing the MySQL part.

YC:
It's possible to use service containers when it comes to spawning dbs. 
https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers
There wouldn't be any teardown required if test data is imported/populated on each execution.

PR #432, opened 2022-12-04
Upgrade testing for SQLite

YC:
How it works:
1. Download release version of Wakapi
2. Run tests against release version
3. Run current build of wakapi
4. If the service can come up, it means that migrations succeeded

This PR also changes `run_api_tests.sh` to have two optional arguments database type and upgrade testing, which defaults to sqlite and no upgrade testing respectively. 

Part of #407

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,4 +117,4 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Run sqlite
-      run: ./testing/run_api_tests.sh sqlite y
+      run: ./testing/run_api_tests.sh sqlite --migration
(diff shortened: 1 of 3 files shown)


muety:
Great contribution! :pray: A few remarks:
* [ ] Can you please add a `wait_for_wakapi` statement before the first test run?
* [ ] Could you add another 1-2 seconds sleep to the wait function? I noticed that even when the instance is up, the first one or two test cases still fail due to connection reset
* [ ] If one of the tests fails, the Wakapi process is not properly terminated, but remains running and thus blocks the port
* [ ] The second positional argument `"y"` looks a bit non-intuitive for an outside user. Could you make it `--migrations` or so?
* [ ] Minor one: add logging before the release downloading
* [ ] Minor one: lower-case variable names (for consistency)

muety:
Awesome contribution, thanks you so much!

PR #436, opened 2022-12-06
fix: disabling tcp webserver sockets, fixes #434

### LINKED ISSUE #434: Cannot disable listing on ports
**Describe the bug**
I disabled like described in the README the listing for ipv4 and ipv6 but they still listen

https://github.com/shyim/nixcfg/blob/0bfb2d62d0d3bfbd6b7730938a33524b0669b963/systems/shea/services/wakapi.nix#L16-L18

**System information**
* Wakapi version
* Operating system
    * If Linux: which distro?
    * If Docker: which image and tag?
* Database (SQLite, MySQL, ... ?)

shyim:
When you pass an empty string, the default will be applied from the config struct

https://github.com/muety/wakapi/blob/master/config/config.go#L114

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,8 +4,6 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"github.com/muety/wakapi/utils"
-	"github.com/robfig/cron/v3"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -13,6 +11,9 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/muety/wakapi/utils"
+	"github.com/robfig/cron/v3"
 
 	"github.com/emvi/logbuch"
 	"github.com/gorilla/securecookie"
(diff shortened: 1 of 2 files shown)


shyim:
The import changes happend due vscode go plugin idk? Is there any kind of config for this repo?

YC:
I think "" will have to be retained alongside "-". 
Would be a breaking change otherwise.

shyim:
Which `""` you mean?

YC:
> Which `""` you mean?

Previously, if I had a `config.yml`:
```yaml
listen_ipv4: 127.0.0.1              # leave blank to disable ipv4
listen_ipv6:                               # leave blank to disable ipv6
```
This would disable IPv6. 
However, given that the code is now looking for "-", this no longer works unless you update the config to `listen_ipv6: "-"`. 

It may be better to redesign this and just keep one field for address, but that'll probably also be a breaking change.

shyim:
I could also check for both in the IF and log something when empty string is used?

shyim:
changed :)

muety:
Closes #434

PR #437, opened 2022-12-06
feat: allow using mysql socket, fixes #433

### LINKED ISSUE #433: Allow connecting using MySQL socket
This line forces the client to use TCP https://github.com/muety/wakapi/blob/master/config/db.go#L57

It looks like you can do also unix see https://stackoverflow.com/questions/67867424/how-to-connect-with-mysql-using-a-unix-socket

shyim:


shyim:
Could you maybe create a new tag? :) I would like to get rid of my patch and maybe publish this as package at nixpkgs :)

muety:
New release is coming soon! Hopefully even still this week. Stay tuned!

PR #438, opened 2022-12-06
fix: remove config file requirement, fixes #435

### LINKED ISSUE #435: Allow to run without any config.yml
**Describe the bug**
Currently, wakapi throws a panic when it cannot find a config.yml. You can specify any configuration using Environment. It's not necessary to required to have an config.yml

shyim:
configor does not throw an error when the file is not there :)


muety:
Thanks a lot for all your recent contributions! :pray:

PR #448, opened 2022-12-27
Basic migration tests for mysql/mariadb/postgres

YC:
Part of #407


muety:
Thank you @YC, such a great contribution!! Only a few remarks (comments on the code)...

testing/run_api_tests.sh
```diff
@@ -91,5 +122,9 @@ if [[ $MIGRATION -eq 1 ]]; then
     kill -TERM $pid
 fi
 
+if [ "$docker_down" -eq 1 ]; then
```
muety, line 125: Tests run through, but I'm still getting this error (e.g. when running `bash testing/run_api_tests.sh sqlite`).

```
[...]
Shutting down Wakapi ...
testing/run_api_tests.sh: line 125: [: : integer expression expected
Exiting with status 0
```

testing/run_api_tests.sh
```diff
@@ -37,6 +36,33 @@ fi
 
 # Initialise test data
 case $1 in
+    postgres|mysql|mariadb)
+    docker compose -f "$script_dir/docker-compose.yml" down
+    docker volume rm "testing_wakapi-$1"
+
+    docker_down=1
+    docker compose -f "$script_dir/docker-compose.yml" up --wait -d "$1"
```
muety, line 44: I think this is supposed to read `docker-compose`, no?

testing/run_api_tests.sh
```diff
@@ -37,6 +36,33 @@ fi
...
+    if [ "$1" == "mariadb" ]; then
+        config="config.mysql.yml"
+    else
+        config="config.$1.yml"
+    fi
+
+    db_port=0
+    if [ "$1" == "postgres" ]; then
+        db_port=5432
```
muety, line 53: I'd probably use some different, less common port (e.g. `55432` or so). This way, tests could still run, even if a Postgres database (/ MySQL, respectively) is already running on the machine for a different purpose.

testing/run_api_tests.sh
```diff
@@ -91,5 +122,9 @@ if [[ $MIGRATION -eq 1 ]]; then
     kill -TERM $pid
 fi
 
+if [ "$docker_down" -eq 1 ]; then
+    docker compose -f "$script_dir/docker-compose.yml" down
+fi
+
 echo "Exiting with status $exit_code"
 exit $exit_code
```
muety, line 130: During one test, I still encountered that issue where the testing Wakapi instance was still running in the background and blocking the port when there had been a failure during the tests. Couldn't really reproduce that again, though. So maybe quickly revise the "killing" logic again, if you can find a clue on how that happened, but otherwise I think it's fine :+1:.

muety (changes requested):

testing/run_api_tests.sh
YC: Looks like this won't be possible. 
https://github.com/muety/wakapi/actions/runs/3786548513/jobs/6437545137#step:4:111

testing/run_api_tests.sh
YC: Should be addressed by default value, if not defined.

testing/run_api_tests.sh
YC: Here's an explanation of the new syntax:
https://stackoverflow.com/a/66516826

testing/run_api_tests.sh
YC: Revised. Should be somewhat more robust.

testing/run_api_tests.sh
muety: Note to all Linux users: you'll need to [install `docker-compose-plugin`](https://docs.docker.com/compose/install/linux/#install-using-the-repository).

muety:
Just realized the migration script has a little flaw. When a release for a new version is already published, while the CI pipeline for that version only starts running, the script will compare the version against itself, instead of with the previous release. We could add some logic to use the second-last release if two version tags are the same or something. But I think it's not worth the effort. We'll just have to keep in mind to only publish a new release _after_ CI has run through fine (should be obvious anyway).

muety:
Just randomly came across https://github.com/ory/dockertest. Looks quite useful! Probably not particularly for our use cases, but probably worth keeping it in mind in general with regard to testing.

PR #456, opened 2023-01-16
fix(metrics): warning - postgres - failed to get database size (expected 0 arguments, got 1)

Kichiyaki:
Hello, I've just enabled metrics and noticed this warning, so I've prepared a fix.


muety:
Thanks a lot!

PR #458, opened 2023-01-19
fix: when importing from wakapi, os & editor are reversed

xxchan:
image


muety:
Oh wow, I'm surprised nobody (including myself) ever noticed that before! Thanks for this contribution :+1:

PR #461, opened 2023-01-21
test: remove api call for timezone

YC:
Removes API call in Postman API Tests. 

```js
// Adapted from https://stackoverflow.com/a/63199512
function getTimezoneOffset(timeZone, date = new Date()) {
  const tz = date.toLocaleString("en", {timeZone, timeStyle: "long"}).split(" ").slice(-1)[0];
  const dateString = date.toString();
  const offset = Date.parse(`${dateString} UTC`) - Date.parse(`${dateString} ${tz}`);
  return offset / 1000 / 60;
} 
// getTimezoneOffset("Europe/Berlin") == 60
// getTimezoneOffset("Europe/Berlin", new Date("2022-04-01")) == 120
```


YC:
@muety Any plans to move away from `gorilla/mux`, given that it's deprecated?
I think the Verb Tampering is due to redirect behavior in `mux`.

muety:
Another great contribution, thanks a lot! :pray: 

> Any plans to move away from `gorilla/mux`, given that it's deprecated?

I really liked Gorilla for being very modular and composable. For example, you could choose to only use the router, or also plug in some middleware, session handling, etc. and everything worked perfectly well together. I wouldn't want to use a fully-fledged web framework like Gin or Chi or so, but keep things rather minimalist. Now that Gorilla is deprecated, we'll unfortunately have to find a replacement for it in the medium-long run. But I didn't really do a lot of research on possible alternatives, yet. We're currently using:

* `gorilla/mux` (router)
* `gorilla/securecookie` (encrypted cookies)
* `gorilla/session` (cookie-based session handling)
* Some parts of `gorilla/handlers` (different middlewares)

We would have to find good replacements for each of these, which are compatible among each other and which, ideally, don't require to commit on a whole framework. If you have anything in mind, please let me know!

> I think the Verb Tampering is due to redirect behavior in `mux`.

It's related to `mux`'s [`StrictSlash`](https://pkg.go.dev/github.com/gorilla/mux#Router.StrictSlash) behavior. But I don't think it's a problem for us. When requesting `DELETE /api//health`, you'll get a `301` redirect to `/api/health`. Most clients will perform that follow-up request as a `GET`, so all good. But also if they did a `DELETE` in the second request, they'd end up with a `405` then. So in my understanding, there is no security risk related to this. Correct me if I'm wrong.

PR #462, opened 2023-01-24
Bulk label association

### LINKED ISSUE #368: Feature request: Add label to many projects at once
I really like the project label feature, and I use it quite a bit:

image

I recently started working at a new company, which means working with more than 25+ new repositories within the last months. I want to label these projects as "Company" but doing so one-by-one is quite tedious.

Simple QOL request, it would be awesome if we could do something like:

wakapi-projects-picker

That way, you can do it once instead of picking Project / Label and hitting save 25 times

IgnisDa:
Fixes #368.


IgnisDa:
Here is what it looks like till now

image

IgnisDa:
Bulk association is also complete. Here is what it looks like:

image

routes/settings.go
```diff
@@ -364,22 +365,37 @@ func (h *SettingsHandler) actionAddLabel(w http.ResponseWriter, r *http.Request)
...
-	label := &models.ProjectLabel{
-		UserID:     user.ID,
-		ProjectKey: r.PostFormValue("key"),
-		Label:      r.PostFormValue("value"),
-	}
+	var labels []*models.ProjectLabel
 
-	if !label.IsValid() {
-		return http.StatusBadRequest, "", "invalid input"
+	if r.PostFormValue("num_projects") == "multiple" {
```
muety, line 370: I think you don't even need this distinction here. You can probably just do `<select name="key">` and both will get parsed as a list, while in the first case (non-batch associate), it will only have one entry.

muety:
Wow, great contribution, thanks a lot! Only a few, minor, UI-related change requests
* [ ] In the "single-add" case, can you make either select and the button a bit more subtle, or potentially only appear after clicking a very subtle, gray "+" icon or so? With many labels and many projects, it looks a bit too cluttered with an additional select box and additional big green button each time. Or, alternatively, at least make the button a bit smaller (e.g. only a "+") and both elements (select and button) a little bit more offset to the right from the actual labels.
* [ ] In the multi-select, could you make the options appear dark gray instead of white when they loose focus?

muety (changes requested):
Wow, great contribution, thanks a lot! Only a few, minor, UI-related change requests
* [ ] In the "single-add" case, can you make either select and the button a bit more subtle, or potentially only appear after clicking a very subtle, gray "+" icon or so? With many labels and many projects, it looks a bit too cluttered with an additional select box and additional big green button each time. Or, alternatively, at least make the button a bit smaller (e.g. only a "+") and both elements (select and button) a little bit more offset to the right from the actual labels.
* [ ] In the multi-select, could you make the options appear dark gray instead of white when they loose focus?

IgnisDa:
Here is the new UI.

image

muety:
Thank you! :+1:

IgnisDa:
@muety Could you release the docker image on ghcr as well?

muety:
Wanna wait for a few more changes / fixes to accumulate before next release, if you don't mind.

notarock:
This is awesome, thank you!

PR #465, opened 2023-01-25
Non-root user for Docker

### LINKED ISSUE #464: Don't run as root user in Docker
https://sonarcloud.io/project/security_hotspots?id=muety_wakapi&hotspots=AYXobH5SMTp39knQix3p

Instead, create an `app` user or so and use that to run Wakapi.

YC:
Resolves #464 
https://github.com/nodejs/docker-node/blob/main/19/alpine3.17/Dockerfile#L5-L6
https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,6 @@ Dockerfile
 docker-compose.yml
 .dockerignore
 .git*
+node_modules/
+testing/*.zip
+testing/wakapi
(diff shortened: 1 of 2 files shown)


muety:
Wow, that was quick! :clap:

Issue #408, opened 2022-09-08
👉 Replace SQLite with pure Go implementation (need your help!) 👈


muety:
Wakapi uses SQLite as its default database via [mattn/go-sqlite3](https://github.com/mattn/go-sqlite3) and the official GORM driver [go-gorm/sqlite](https://github.com/go-gorm/sqlite). SQLite is implemented in C, so if used in a project, the project must be compiled with `CGO_ENABLED=1`. That, in turn, requires to have a C-compiler (like `gcc` or `mingw`) installed and generally caused a bunch of problems in the past (e.g. when it comes to cross-compilation). 

For quite a while, I was looking out for an implementation of SQLite that is in pure Go - so we can get rid of _cgo_ - and eventually found [glebarez/go-sqlite](https://github.com/glebarez/go-sqlite), which is based on [cznic/sqlite](https://gitlab.com/cznic/sqlite), plus [glebarez/sqlite](https://github.com/glebarez/sqlite) as a compatible GORM driver (relationship among these packages explained [here](https://github.com/go-gorm/gorm/issues/4101#issuecomment-1241060910)). The base package [is actually a transpiled version](https://gitlab.com/cznic/sqlite/-/issues/107) of the original C code. Note that _mattn/go-sqlite3_ is still included as a dependency, but apparently only for testing (or so?), so you can well compile without _cgo_.

The package [claims to be fully compatible](https://gitlab.com/cznic/sqlite/-/issues/107) with the reference implementation and able to act as a drop-in replacement, but comes with a downside in performance (see [benchmarks](https://github.com/glebarez/go-sqlite/tree/master/benchmark)). I'd assume that you wouldn't recognize the worse performance with average Wakapi setups, though.

Since replacing the database implementation is quite a big change, **I need your help here**! If you're currently self-hosting Wakapi with an SQLite database, it would be super helpful if you could test the new, C-free version as part of your setup. Backup your database (just in case), replace the original Wakapi binary with the one below, keep using Wakapi for a couple of days, see if any errors occur and report them back. Thank you! :pray:

* :point_right: **Windows:** [wakapi_windows_amd64.zip](https://github.com/muety/wakapi/files/9529995/wakapi_windows_amd64.zip)
* :point_right: **Linux:** [wakapi_linux_amd64.zip](https://github.com/muety/wakapi/files/9529993/wakapi_linux_amd64.zip)

Compiled locally, doing:
```bash
[...]

PR #501, opened 2023-07-09
ci(release): update to ubuntu-latest

YC:
ubuntu-18.04 has been deprecated.
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

I think we downgraded to 18.04 at some point due to glibc, but this is perhaps not relevant any more after #408?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,10 +12,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - platform: ubuntu-18.04
+          - platform: ubuntu-latest
             GOOS: linux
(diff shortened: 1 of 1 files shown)


YC:
@muety I've attached the artifacts from:
https://github.com/muety/wakapi/actions/runs/5497886373
As release assets for 2.8.0.

muety:
Awesome, thank you!

PR #503, opened 2023-07-13
fix(import): data dump already exists

### LINKED ISSUE #502: Import from Wakatime fails when heartbeats data dump already exists
**Describe the bug**

I created my instance of wakapi, but restarted the instance during the import, which probably interrupted the data import. Now when I try to start it again, the UI says the import is being started, but in server log I can see that Wakatime refused to create the data dump with 400 Bad Request response. When I try to create the dump from wakatime frontend, I get the same error, with more detail: `Wait for your current export to expire before creating another.`

Wakapi should therefore list current data dumps before attempting to create a new one.

**System information**
* Wakapi version: latest docker image (2.8.1)

WangEdward:
handle the import when there is already an active data dump.

Resolves #502

### DIFF
--- a/services/imports/wakatime_dump.go
+++ b/services/imports/wakatime_dump.go
@@ -48,7 +48,7 @@ func (w *WakatimeDumpImporter) Import(user *models.User, minFrom time.Time, maxT
 			if err := json.NewDecoder(res.Body).Decode(&datadumpError); err != nil {
 				return nil, err
 			}
-			if datadumpError.Error == "Wait for your current export to expire before creating another."{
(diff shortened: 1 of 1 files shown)


muety:
Thanks for this contribution. I made the code / logic even a bit simpler and opened #504. Please have a look if that works for you.

Also, next time please respect it when an issue already has an assignee.

muety:
You could have left it open, I based my changes on your commits, so this one would have gotten merged as well. But either way, changes are in! Thanks again for the contribution.

PR #504, opened 2023-07-13
Simplify import checks

muety:
Adds on #503. @WangEdward, please check.


WangEdward:
Thank you! Your way of handling this is much better than mine 🤩

Sorry for not noticing the assignee 🥺 I will pay attention to it next time!

I will close #503 then.

Issue #33, opened 2020-06-06
OAuth 2 login


muety:
Log in with:
* GitHub
* GitLab
* ... ?

muety:
* https://godoc.org/github.com/markbates/goth
* https://godoc.org/github.com/markbates/goth/gothic

alexanderadam:
It would be nice to use external auth services like [Authelia](https://www.authelia.com/), [Keycloak](https://www.keycloak.org/), [Authentik](https://goauthentik.io/) or [GitLab](https://docs.gitlab.com/ee/integration/openid_connect_provider.html).

And I guess OpenID Connect (OIDC) /OAuth2 is pretty much standardized and supports, so this could be a good solution.
Most projects allow to set the URLs and secrets via ENV variables. This way users have the full flexibility what kind of provider they want to use.

muety:
See https://github.com/muety/wakapi/issues/94#issuecomment-1366855621 for additional thoughts on this.

theredditbandit:
Hi @muety,

I read the discussion surrounding this issue and saw that you mentioned in [a comment](https://github.com/muety/wakapi/issues/94#issuecomment-1367196193) on #94:
> OAuth 2 / OIDC authentication for Wakapi itself is planned as well, see https://github.com/muety/wakapi/issues/33. I didn't get to implementing that, yet, because it's quite a big one (requires refactorings to current auth. logic). But I see how it will be a huge benefit for users, so please stay tuned!

I'd love to contribute to implementing the OAuth 2 / OIDC authentication feature. Would you be open to accepting a pull request for this enhancement?

To help me get started, could you provide some guidance on the specific areas of the authentication logic that need refactoring? I'd appreciate any pointers to relevant files or architectural considerations you have in mind.

PR #505, opened 2023-07-14
feat: add OAuth support for Wakapi

WangEdward:
TODOs:
- [x] add config for OAuth
- [x] add api interface for OAuth
- [ ]  login and bind existing user
  - [ ] database 
  - [ ] login logic
  - [ ] front end

#33

### DIFF
--- a/config.default.yml
+++ b/config.default.yml
@@ -74,6 +74,12 @@ subscriptions:
   stripe_endpoint_secret:
   standard_price_id:
 
+oauth:
+  enabled: false                        # whether to enable oauth
+  provider:                             # one of ['github', 'microsoft', 'google']
+  client_id:                            # oauth client id
+  client_secret:                        # oauth client secret
+
 mail:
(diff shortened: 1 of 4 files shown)


muety:
Thanks a lot for all the effort you must have put into this! The more I feel bad when unfortunately having to reject this, sorry!

I wouldn't want to implement security-related things like these ourselves. I see no point in manually implementing logic to do per-provider-specific requests, callbacks, URL parameter parsing, etc. when there are already existing libraries for this (e.g. see [goth](https://godoc.org/github.com/markbates/goth) / [gothic](https://godoc.org/github.com/markbates/goth/gothic), linked in the original issue above), that give you a ton of built-on providers, plus properly tested and robust implementations for free.

Goth looks like a good solution at first sight. And maybe there are even more of the same kind. Doing a bit of research on this would be interesting and then proposing a "concept" how exactly this would fit into our user system / user dabase.

muety:
Btw., very minor one, but: by convention you'd use camel-case variable names in Go, no snake-case.

PR #507, opened 2023-07-20
fix failing migration which prohibits startup

cbrand:
Fix an issue in migration which results in the following error message due to wrong and or precedence configuration:
```
panic: runtime error: index out of range [0] with length 0
```


muety:
Thank you!

PR #528, opened 2023-09-24
ci: reduce permissions for non-release actions

YC:
See "Use credentials that are minimally scoped" in https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

Note that `GITHUB_TOKEN` does require more privileges for release actions (e.g. for pushing docker image). 
However, there's probably less risk there given that those pipelines are only triggered on release.


muety:
Awesome, thank you!

PR #547, opened 2023-11-13
fix: ignore permissions when user is logged in stats api

### LINKED ISSUE #546: Stats endpoint no data when logged in
**Describe the bug**
If I call the `api/compat/wakatime/v1/users/current/stats/last_7_days` endpoint I can't see any data regarding editors, machines, projects, operating_systems. This is because I have disabled their switch in the permission tab for public access. The issue is that this occurs **even when I'm logged in**. If I enable the switches I can correctly see the data.

The stats endpoint also lacks of the [best_day](https://wakatime.com/developers/#stats) propriety, although this may be a separate issue. 

**System information**
* Wakapi version: v2.9.3
* Operating system: Debian 11
* Database: mysql

GioPan04:
Fixes #546

### DIFF
--- a/routes/compat/wakatime/v1/stats.go
+++ b/routes/compat/wakatime/v1/stats.go
@@ -1,10 +1,11 @@
 package v1
 
 import (
-	"github.com/go-chi/chi/v5"
-	"github.com/muety/wakapi/helpers"
 	"net/http"
 	"time"
+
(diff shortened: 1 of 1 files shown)


muety:
Awesome, thanks your contribution! Change will be included in the next release.

PR #550, opened 2023-11-15
feat: added fields in projects api

GioPan04:
Added fields in Project model for WakaTime compat API:
 - `last_heartbeat_at`
 - `human_readable_last_heartbeat_at`
 - `urlencoded_name`
 - `created_at`

I removed the `repository` field because I saw that there is no implementation. It used to return an empty string, which does not reflects the WakaTime implementation, because it's either null or an object. After #33 gets solved, I think it could be added again.

### DIFF
--- a/models/compat/wakatime/v1/project.go
+++ b/models/compat/wakatime/v1/project.go
@@ -1,11 +1,16 @@
 package v1
 
+import "time"
+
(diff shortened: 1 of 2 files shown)


muety:
Another awesome contribution, thanks!

PR #553, opened 2023-11-17
feat: show avatar url in wakatime user model

GioPan04:
- Added the `photo` field in the User WakaTime model. It returns the URL for the avatar API used in the frontend.
 - Replaced "Anonymous User" with the user id (should be more clear) in `display_name`

### DIFF
--- a/models/compat/wakatime/v1/user.go
+++ b/models/compat/wakatime/v1/user.go
@@ -1,8 +1,11 @@
 package v1
 
 import (
-	"github.com/muety/wakapi/models"
+	"fmt"
 	"time"
+
+	"github.com/muety/wakapi/config"
+	"github.com/muety/wakapi/models"
 )
 
 const DefaultWakaUserDisplayName = "Anonymous User"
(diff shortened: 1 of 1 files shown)


models/compat/wakatime/v1/user.go
```diff
@@ -36,14 +40,17 @@ func NewFromUser(user *models.User) *User {
...
 	return &User{
 		ID:          user.ID,
-		DisplayName: DefaultWakaUserDisplayName,
+		DisplayName: user.ID,
 		Email:       user.Email,
 		TimeZone:    tz,
 		Username:    user.ID,
 		CreatedAt:   user.CreatedAt,
 		ModifiedAt:  user.CreatedAt,
+		Photo:       fmt.Sprintf("%s%s/%s", config.Server.GetPublicUrl(), config.Server.BasePath, user.AvatarURL(config.App.AvatarURLTemplate)),
```
GioPan04, line 53: I think that the CI test are failing because the config is not initialized during testing. I never done testing in Go and so I don't know where to fix

PR #556, opened 2023-11-17
fix: run api tests ordering

YC:
Reorder such that:
1. Run old version (initialisation for old db)
2. Run new version (migration test)
4. If sqlite (api tests)
    1. populate data
    2. Run new version again
    3. Run api test

### DIFF
--- a/testing/run_api_tests.sh
+++ b/testing/run_api_tests.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
+set -o nounset -o pipefail -o errexit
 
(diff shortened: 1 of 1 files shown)


muety:
Thank you! :raised_hands: :pray: 

Briefly for my understanding: tests were failing recently, because we tried to run the API tests against the old release, which obviously didn't have the latest changes included, that we asserted in those tests? Correct?

YC:
Yes, that's correct.

PR #561, opened 2023-11-20
fix: avatar url wrong url using external api

GioPan04:
When using an external API service (like Gravatar) the avatar URL returned from `/api/compat/wakatime/v1/users/current` is wrong: `http://localhost:3000/https://gravatar.com/avatar/3c90a086090c74081892efea2f523658`.

This PR checks if the returned URL from User.AvatarURL() is an URL or a path.

### DIFF
--- a/models/compat/wakatime/v1/user.go
+++ b/models/compat/wakatime/v1/user.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


GioPan04:
In my previous PR (#553) I didn't know that using an external avatar service was a thing. Sorry 😅

PR #564, opened 2023-11-24
allow empty version in user agent regex for KDE Kate

InsanePrawn:
My KDE Kate installation with the officially endorsed plugin generates UA strings like `wakatime/ (Linux-6.XX-HOSTNAME-x86_64) KTextEditor/5.111.0 kate-wakatime/1.3.10`


muety:
Thanks for the contribution! This actually looks like a small bug on the plugin's end (perhaps file an issue?). But will include this change anyway for now. Will be published with the next release.

PR #565, opened 2023-11-24
perf: avoid allocations with `(*regexp.Regexp).MatchString`

Juneezee:
We should use `(*regexp.Regexp).MatchString` instead of `(*regexp.Regexp).Match([]byte(...))` when matching string to avoid unnecessary `[]byte` conversions and reduce allocations. A one-line change for free performance improvement.

Example benchmark:

```go
func BenchmarkMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := emailAddrRegex.Match([]byte("email@example.com")); !match {
			b.Fail()
		}
	}
}

func BenchmarkMatchString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := emailAddrRegex.MatchString("email@example.com"); !match {
			b.Fail()
		}
	}
}
```

Result:

```
goos: linux
goarch: amd64
pkg: github.com/muety/wakapi/models
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkMatch-16          	 1149106	      1012 ns/op	      24 B/op	       1 allocs/op
BenchmarkMatchString-16    	 1566058	       727.8 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/muety/wakapi/models	3.521s
```

### DIFF
--- a/models/mail_address.go
+++ b/models/mail_address.go
@@ -48,7 +48,7 @@ func (m MailAddress) Domain() string {
 }
 
 func (m MailAddress) Valid() bool {
-	return emailAddrRegex.Match([]byte(m))
+	return emailAddrRegex.MatchString(string(m))
 }
 
 func (m MailAddresses) Strings() []string {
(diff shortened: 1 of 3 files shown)


muety:
Thanks 👍 This change won't actually have a real effect, since those functions are not used frequently. But still good to have.

PR #582, opened 2023-12-23
Updated SALT command syntax

SupianIDz:
Screenshot 2023-12-24 at 03 06 17

In MacOS Sonoma error appears that the `tr` command is encountering some bytes in the `/dev/urandom` output that are not valid. The `/dev/urandom` device produces random bytes.
When you pipe this to tr, it expects input that is valid text in your current locale. However, urandom may produce bytes that are not valid.

The `LC_ALL=C` sets the locale to C, which handles any byte values. Now should not complain about illegal byte sequences anymore. This will generate a random alphanumeric string.

Tested on MacOS and Ubuntu. In the left picture is MacOS, the right is Ubuntu
Screenshot 2023-12-24 at 03 09 29

So this doesn't break anything

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ $ eget muety/wakapi
 # Create a persistent volume
 $ docker volume create wakapi-data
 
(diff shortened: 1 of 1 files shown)


muety:
Thanks for spotting this :+1:

PR #592, opened 2024-01-10
Add mssql support

### LINKED ISSUE #591: [Feature request] Support SQL Server as database
It would be great if wakapi can support SQL Server as the database. [gorm does support sql server](https://gorm.io/docs/connecting_to_the_database.html#SQL-Server), so it shouldn't be so hard to support it. I am already working on it [here](https://github.com/ddadaal/wakapi-sqlserver/tree/add-mssql-support).

Blockers:

-  SQL Server doesn't allow multiple columns in a table with `timestamp` type

Reason: [this stackoverflow question](https://stackoverflow.com/questions/12063850/multiple-timestamp-columns-in-sql-server-2000)

Workaround: Use [gorm's support for custom data type](https://gorm.io/docs/data_types.html#GormDataTypeInterface). Implemented `GormDBDataType` interface for `CustomTime`, use `datetimeoffset` for sqlserver dialector.

It adds a new optional `TIMESCALE` gorm tag, to set the scale (precision) of the `datetimeoffset` or `timestamp`. For example, `timeScale:3` yields `datetimeoffset(3)` for sql server, and `timestamp(3)` for other dialector.

- `./testing/run_api_tests.sh sqlserver --migration` cannot connect to db, but `./testing/run_api_tests.sh sqlserver` can.

Still investigation.

- `Could not create constraint or index. See previous errors.`

Still under investigation 

- `default:CURRENT_TIMESTAMP` might not be optimal for `doptimaloffset`

ddadaal:
image

Fixes #591.

Notes on MSSQL:

- no TRUE or FALSE literals in mssql.
    - Use 1 and 0 instead
- cannot give an alias to the table when update: in `update table as t set xxx = xxx`, `as t` is invalid
- mssql doesn't allow circular cascade paths on foreign keys (see [this stackoverflow answer](https://stackoverflow.com/a/852047))
- some words are not allowed as identifiers (like `users`)
    -  Add a `utils.QuoteDbIdentifiers` and `utils.QuoteSql` functions to quote identifiers in SQLs in any dialectors
- use `datetimeoffset` instead of `timestamp` for datetime when modeling. 
    - Use `CustomTime` for datetime properties/columns and don't specify type using tag. `CustomTime` will choose correct type for all dbs
- no `join using`
    - use regular `join on` instead

Tested functionalities:

- [x] Login and Register
- [x] Collect data from wakatime plugins
- [x]  Dashboard
- [x] Projects
- [x] Leaderboards
- [ ] Email (not data yet)
- [ ] Integration with wakatime 
    - [x] sync to wakatime.com
    - [ ] import data (import can start, but no data can be downloaded. error is `Get "https://api.wakatime.com/api/v1/users/current/data_dumps": unexpected EOF`. seems not related to db.


repositories/user.go
```diff
@@ -85,12 +86,12 @@ func (r *UserRepository) GetByLoggedInAfter(t time.Time) ([]*models.User, error)
...
+		Select("user_id , max(time) as time").
 		Group("user_id")
 
 	var userIds []string
 	if err := r.db.
-		Select("user as id").
+		Select("user").
 		Table("(?) as q", subQuery1).
 		Where("time >= ?", t.Local()).
 		Scan(&userIds).Error; err != nil {
```
muety, line 97: These changes break the `GetByIds` function. Getting a 500 now when request, for example, the `/api/metrics` endpoint with SQLite:

```
2024-01-12T15:29:56.148722234+01:00 [ERROR] failed to retrieve active users for metric - SQL logic error: no such column: user (1)
2024-01-12T15:29:56.148733462+01:00 [ERROR] SQL logic error: no such column: user (1)
```

repositories/heartbeat.go
```diff
@@ -231,22 +233,29 @@ func (r *HeartbeatRepository) GetUserProjectStats(user *models.User, from, to ti
...
-			"select project, user_id, min(time) as first, max(time) as last, count(*) as cnt "+
+			"select project as p, user_id, min(time) as first, max(time) as last, count(*) as cnt "+
 			"from heartbeats "+
 			"where user_id = ? and project != '' "+
 			"and time between ? and ? "+
 			"and language is not null and language != '' and project != '' "+
 			"group by project, user_id "+
 			"order by last desc "+
-			"limit ? offset ? "+
+			"offset ? "+limitClause+
```
muety, line 252: Getting an error when calling an endpoint that calls down to this function.

```
2024-01-12T15:34:52.198713355+01:00 [ERROR] SQL logic error: near "offset": syntax error (1)
```

muety (changes requested):
Thanks a lot for this contribution and for diving into the depths of all of this! :raised_hands: I did a bunch of quick tests myself with MSSQL and from what I found, everything is working like a charm!

Only when running the API tests against SQLite, I'm getting a bunch of errors, including 500 responses for the metrics- and WakaTime projects-endpoints.

models/shared.go
```diff
@@ -35,6 +39,23 @@ type KeyedInterval struct {
...
+
+	t := "timestamp"
+
+	if db.Config.Dialector.Name() == (sqlserver.Dialector{}).Name() {
+		t = "datetimeoffset"
+	}
+
+	if scale, ok := field.TagSettings["TIMESCALE"]; ok {
+		t += fmt.Sprintf("(%s)", scale)
+	}
```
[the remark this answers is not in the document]
ddadaal: Yes. when gorm parses tags, it converts tags to upper case. I thought gorm would provide a function to query tags under  gorm, but I didn't find it.

https://github.com/go-gorm/gorm/blob/0123dd45094295fade41e13550cd305eb5e3a848/schema/utils.go#L35

repositories/user.go
ddadaal: Yes. This is because the `user` is `select`ed from the subquery above. The `user_id` column was previously given a alias `user`. But the alias was wrongly removed. I am fixing it in the next commit.

repositories/heartbeat.go
ddadaal: It seems because SQLite has to put `limit` before `offset`. I am fixing it so the new clause only applies to mssql.

utils/db.go
```diff
@@ -49,3 +51,37 @@ func WithPaging(query *gorm.DB, limit, skip int) *gorm.DB {
...
+func (s stringWriter) WriteByte(c byte) error {
+	return s.Builder.WriteByte(c)
+}
+
+func (s stringWriter) WriteString(str string) (int, error) {
+	return s.Builder.WriteString(str)
+}
+
+// QuoteDbIdentifier quotes a column name used in a query.
+func QuoteDbIdentifier(db *gorm.DB, identifier string) string {
```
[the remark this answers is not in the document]
ddadaal: Thanks, and the new `QuoteSql` should be better since it can quote multiple args at one function call 😊

muety:
Let me know when you want me to go for another round of review!

ddadaal:
> Let me know when you want me to go for another round of review!

Thanks for your review. Definitely! I focused only on the mssql functionalities in the last few days. I will try with other dbs and fix the problems. Some problems seem to be pretty annoying, especially the foreign key constraint one. It may take a few days to fix all of these.

models/summary.go
```diff
@@ -27,16 +28,20 @@ const DefaultProjectLabel = "default"
...
-	ToTime           CustomTime   `json:"to" gorm:"not null; type:timestamp; default:CURRENT_TIMESTAMP; index:idx_time_summary_user" swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
+	ID       uint       `json:"-" gorm:"primary_key; size:32"`
+	User     *User      `json:"-" gorm:"not null; constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
+	UserID   string     `json:"user_id" gorm:"not null; index:idx_time_summary_user"`
+	FromTime CustomTime `json:"from" gorm:"not null; default:CURRENT_TIMESTAMP; index:idx_time_summary_user" swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
+	ToTime   CustomTime `json:"to" gorm:"not null; default:CURRENT_TIMESTAMP; index:idx_time_summary_user" swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
+	// Previously, all of the following properties created a foreign key constraint on the summary_items table back to this summary table
+	// All of these created foreign key constraints are the same
+	// But mssql will complains about circular cascades on update/delete between these two tables
+	// So, creating one constraint should be enough
```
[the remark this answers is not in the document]
ddadaal: It's fixed now. Adding more comments in the code

repositories/heartbeat.go
```diff
@@ -231,22 +233,29 @@ func (r *HeartbeatRepository) GetUserProjectStats(user *models.User, from, to ti
...
 			") "+
 			"select distinct project, min(first) as first, min(last) as last, min(cnt) as count, first_value(language) over (partition by project order by count(*) desc) as top_language "+
 			"from heartbeats "+
-			"inner join projects using (project, user_id) "+
+			"inner join projects on heartbeats.project = projects.p and heartbeats.user_id = projects.user_id "+
 			"group by project, language "+
-			"order by last desc", user.ID, from, to, limit, offset).
+			"order by last desc", user.ID, from.Format(time.RFC3339), to.Format(time.RFC3339), offset, limit).
```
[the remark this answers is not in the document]
ddadaal: Yes. I just tested `2024-01-13T08:03:26+08:00` and `2024-01-13T08:03:26Z` in MySQL, MSSQL, PostgresSQL and SQLite and all of them support it.

ddadaal:
Hello, I think it's now ready for another review.

muety:
Very cool, thank you so much! Only found one last issue.

Every heartbeat has a `hash` property that is used for detecting duplicates. The heartbeats table has a unique index defined on the according column and in the repository, we effectively set `ON CONFLICT DO NOTHING` (see [here](https://github.com/muety/wakapi/blob/e70cc59528323cd032b1fb6094e65a5fa847ce45/repositories/heartbeat.go#L32)) when inserting new heartbeats. This causes duplicate heartbeats to be silently ignored and prevents from clutter in the database. However, this mechanism doesn't seem to work with MSSQL. 

Try inserting the same heartbeat twice in a row:

```bash
curl --location 'http://localhost:3000/api/heartbeat' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer <YOUR-BASE64-HASHED-API-KEY>' \
--data '[{
    "entity": "/home/user1/dev/project1/main.go",
    "project": "wakapi",
    "language": "Go",
    "is_write": true,
    "type": "file",
    "category": null,
    "branch": null,
    "time": 1705342578.955
}]'
```

Both requests should result in a `201` response status, while only heartbeat entry is actually inserted to the database. When running with MSSQL, I'm getting a `500` error for the second request, though, alongside the following error message:

```
[ERROR] failed to batch-insert heartbeats - mssql: Cannot insert duplicate key row in object 'dbo.heartbeats' with unique index 'idx_heartbeats_hash'. The duplicate key value is (2234263491a774d0)
```

Any ideas how to work around this?

ddadaal:
> Very cool, thank you so much! Only found one last issue.
> 
> Every heartbeat has a `hash` property that is used for detecting duplicates. The heartbeats table has a unique index defined on the according column and in the repository, we effectively set `ON CONFLICT DO NOTHING` (see [here](https://github.com/muety/wakapi/blob/e70cc59528323cd032b1fb6094e65a5fa847ce45/repositories/heartbeat.go#L32)) when inserting new heartbeats. This causes duplicate heartbeats to be silently ignored and prevents from clutter in the database. However, this mechanism doesn't seem to work with MSSQL.
> 
> Try inserting the same heartbeat twice in a row:
> 
> ```shell
> curl --location 'http://localhost:3000/api/heartbeat' \
> --header 'Content-Type: application/json' \
> --header 'Authorization: Bearer <YOUR-BASE64-HASHED-API-KEY>' \
> --data '[{
>     "entity": "/home/user1/dev/project1/main.go",
>     "project": "wakapi",
>     "language": "Go",
>     "is_write": true,
>     "type": "file",
>     "category": null,
>     "branch": null,
>     "time": 1705342578.955
> }]'
> ```
> 
> Both requests should result in a `201` response status, while only one heartbeat entry is actually inserted to the database. When running with MSSQL, I'm getting a `500` error for the second request, though, alongside the following error message:
> 
> ```
> [ERROR] failed to batch-insert heartbeats - mssql: Cannot insert duplicate key row in object 'dbo.heartbeats' with unique index 'idx_heartbeats_hash'. The duplicate key value is (2234263491a774d0)
> ```
> 
> Any ideas how to work around this?

It is because of a bug in sqlserver dialector. [An issue for it](https://github.com/go-gorm/sqlserver/issues/100) has been opened for a long time, but have not got response from gorm authors yet. I tried to fix it ([attempted fix commit](https://github.com/ddadaal/gorm-sqlserver/commit/b8fb7484bc099b62b153fe2c9aaaa245409cfa3f)) but it didn't work, and the problem seems too complicated. 

I developed a workaround that for mssql, all heartbeats will be inserted one by one, and errors caused by duplicate hash will be ignored. It guarantees correctness but impacts performance. Considering most of time only one heartbeat will be inserted at once, I think the performance penalty should be acceptable.  

Another alternative approach is to write raw sql just for mssql. But mssql doesn't support `on conflict` clause. Although [there are a lot of ways](https://michaeljswart.com/2017/07/sql-server-upsert-patterns-and-antipatterns/) to emulate the behavior, all of them are so complex that I think it is not worth it.

muety:
Thanks a lot for the detailed explanation and for the fix. I think it's very reasonable as a workaround. Again: :pray: :raised_hands:.

muety (approved):

muety:
I'll let the code run in production for a short while and push out a new release soon if no errors pop up (likely tomorrow or so).

models/user.go
```diff
@@ -36,8 +36,8 @@ type User struct {
 	ResetToken          string      `json:"-"`
 	ReportsWeekly       bool        `json:"-" gorm:"default:false; type:bool"`
 	PublicLeaderboard   bool        `json:"-" gorm:"default:false; type:bool"`
-	SubscribedUntil     *CustomTime `json:"-" gorm:"type:timestamp" swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
-	SubscriptionRenewal *CustomTime `json:"-" gorm:"type:timestamp" swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
+	SubscribedUntil     *CustomTime `json:"-" gorm:"swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
+	SubscriptionRenewal *CustomTime `json:"-" gorm:"swaggertype:"string" format:"date" example:"2006-01-02 15:04:05.000"`
```
muety, line 40: @ddadaal Just came across these two lines. The "swaggertype" thing looks odd there. I presume that's not the way how it's supposed to be?

ddadaal: yes...it seems the `gorm:"` part should be removed

muety: :+1: Will do

PR #597, opened 2024-01-24
Fix entity filter

hesstobi:
The dropdown menus to filter for project, language, machine, and label are not working because the selection on the EntityFilter is not updated when selecting a new value. This PR set the selection from the event target.

### DIFF
--- a/static/assets/js/components/entity-filter.js
+++ b/static/assets/js/components/entity-filter.js
@@ -9,7 +9,7 @@ function EntityFilter({type, options, selection}) {
             return this.type.capitalize()
         },
         onSelectionUpdated(e) {
-            this.selection = e.target.value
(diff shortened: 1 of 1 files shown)


muety:
Thanks for this! The filters are working fine in Firefox, I actually never realized they were broken in Chrome. While you're on it, could you please also add functionality to reset the respective filters when going back to _Filter by language ..._ (etc.) in the dropdown? This should revert back to an unfiltered view. Essentially just the way it currently works in Firefox. That would be great!

hesstobi:
Sorry, for only providing half of the fix at the start. Commit https://github.com/muety/wakapi/pull/597/commits/ed8ad01d8279f011dffc5026cf33ae9cdb88d0f3 should do it.

muety:
Great contribution, thanks a lot!

PR #614, opened 2024-03-06
feat: customizable date format

### LINKED ISSUE #517: customizable date format
image

[users using YMD date format](https://en.wikipedia.org/wiki/Date_format_by_country) may have difficulty reading 'Human' date format.

I'm using patch file to work around the issue, but It'd be great to be able to customize date format.
```go
func FormatDateTimeHuman(date time.Time) string {
	return date.Format("2006-01-02 15:04:05 Mon")
}

func FormatDateHuman(date time.Time) string {
	return date.Format("2006-01-02 Mon")
}
```

https://github.com/muety/wakapi/blob/7f2544f003415eef26eda3c5e0778a8e51814486/helpers/date.go#L35-L42

scarf005:
## Summary 

image
image

- fix #517

## Description

Added two new config, `DateFormat` and  `DateTimeFormat` to make `FormatDateTimeHuman` and `FormatDateHuman` customizable.


muety (changes requested):
Thanks :+1: Could you please add a quick validation to check if the provided date can be parsed correctly? If an invalid date format is provided, the app should crash immediately and not just during runtime. The `Load()` function inside `config.go` already contains a couple of similar checks, just add it in there.

muety (approved):

Issue #526, opened 2023-09-22
Unraid: unable to open database file: out of memory (14)


manuel-rw:
**Describe the bug**

When trying to run Wakapi on Unraid, the container will stop immideatly with the following log:

``` 
2023-09-22T22:23:31.244632728+02:00 [ERROR] unable to open database file: out of memory (14)
2023-09-22T22:23:31.244641106+02:00 [ERROR] could not open database
panic: could not open database

goroutine 1 [running]:
github.com/emvi/logbuch.(*StandardFormatter).Pnc(0xc0000e0000?, {0x116b2c5?, 0x116b2c5?}, {0x0?, 0x0?, 0x0?})
        /go/pkg/mod/github.com/emvi/logbuch@v1.2.0/standard_formatter.go:57 +0x3f
github.com/emvi/logbuch.(*Logger).Fatal(0xc0000e0000, {0x116b2c5, 0x17}, {0x0, 0x0, 0x0})
        /go/pkg/mod/github.com/emvi/logbuch@v1.2.0/logger.go:132 +0x77
github.com/emvi/logbuch.Fatal(...)
        /go/pkg/mod/github.com/emvi/logbuch@v1.2.0/funcs.go:55
main.main()
        /src/main.go:145 +0x6d8
2023-09-22T22:23:19.296427774+02:00 [INFO ] disabling data retention policy, keeping data forever
2023-09-22T22:23:19.296436598+02:00 [WARN ] with sqlite, only a single connection is supported
2023-09-22T22:23:19.296448106+02:00 [INFO ] Wakapi 2.8.2
2023-09-22T22:23:19.296450505+02:00 [INFO ] starting with sqlite3 database
```

The generated docker command by dynamix is:
```bash
[...]

PR #649, opened 2024-05-27
Fix permissions of /app folder for sqlite in docker

anisjonischkeit:
I am trying to run sqlite in docker and was getting the same error as shown in issue #526. The app user does not own the /app directory, and so the it cannot create/open the sqlite file that sits in there. This isn't an issue when using postgres/mysql, but it is when you use a file based db.

This PR sets the /app directory's owner to app, so that the running user can modify the sqlite file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ FROM alpine:3
 WORKDIR /app
 
 RUN addgroup -g 1000 app && \
-    adduser -u 1000 -G app -s /bin/sh -D app && \
+    adduser -u 1000 -G app -s /bin/sh -D app && chown app -R /app && \
     apk add --no-cache bash ca-certificates tzdata
 
 # See README.md and config.default.yml for all config options


anisjonischkeit:
Disregard, This looked like it worked, but there was some other environmental thing that made it look like it made a difference. What I ended up having to do, was specifying the user (app) and voluming to the /data directory (which is owned by app)

My full docker-compose.yml for anyone that has the same issue:

```
version: "3.7"

services:
  wakapi:
    image: ghcr.io/muety/wakapi:latest
    ports:
      - 3000:3000
    restart: always
    environment:
      # See README.md and config.default.yml for all config options
      WAKAPI_DB_NAME: "/data/wakapi_db.db"
      WAKAPI_DB_USER: "wakapi"
      WAKAPI_DB_PASSWORD: "YOUR PASSWORD"
      ENVIRONMENT: "prod"
    user: app
    volumes:
      - ./data:/data
```

PR #652, opened 2024-06-20
Improve badge integration sections

notarock:
This pull request add URL encoding and make the two badges a bit more distinct.

When I was looking at the Badge section under integration, I was confused as to why the same badge was being displayed two times, without realizing one was a direct link to the current instance and the other links to shields.io using your instance's data. By making them a bit distinct, it could help remove that confusion when users first look at this section.

Also, by changing the space by a URL-encoded space in the link, it wont breaks the link when a user try to copy the badge link somewhere else.

Before:

image

Copying the badge gives me badge

After:

image

Copying the badge gives me badge


muety:
Great, thank you! :+1:

PR #659, opened 2024-07-11
ci: run mail tests in workflow

### LINKED ISSUE #658: Run mail tests in CI
In https://github.com/muety/wakapi/commit/a5565b12ea11f0ef62214eb43a60b0b24c37a864 I added integration tests for SMTP mail sending using [smtp4dev](https://github.com/rnwood/smtp4dev). Tests are run as part of the standard test suite, but are skipped silently if no _smtp4dev_ endpoint is available. Running the tests requires to bring up a Docker container for _smtp4dev_, which is why currently these tests can only be run manually for simplicity. However, SMTP tests should be configured as part of the GHA workflow, too. Should be quite straightforward and similar to how we run the API tests against different DBs (running in Docker). Perhaps something for you, @YC?

YC:
Resolves #658

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,9 @@ jobs:
         npm -g install newman
         ./testing/run_api_tests.sh
 
+    - name: Mail Tests
+      run: ./testing/run_mail_tests.sh
+
   migration:
     name: Migration tests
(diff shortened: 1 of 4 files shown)


muety:
Awesome contribution, thank you!!

Issue #480, opened 2023-04-12
Migrate from `logbuch` to `slog`


muety:
* For general logging: https://pkg.go.dev/log/slog
* For Sentry integration: https://github.com/samber/slog-sentry
* For request logging: https://github.com/samber/slog-http

Other logging packages considered: 
* ~~https://github.com/uber-go/zap~~
* ~~https://github.com/TheZeroSlave/zapsentry~~

muety:
Btw., interesting talk: https://www.youtube.com/watch?v=tC4Jt3i62ns

muety:
https://www.reddit.com/r/golang/comments/16ebwk4/how_you_are_using_slog_your_experience_pls/

muety:
Alternatively: https://github.com/rs/zerolog

finnng:
Hi @muety, I can help work on this issue. 
I'm going to use the packages as you mentioned.
- https://pkg.go.dev/log/slog
- https://github.com/samber/slog-sentry
The ticket was created a year ago. Do you have any more suggestions?

muety:
As we're trying to use as few external dependencies as possible in Wakapi, I'd vote for building the structured logging part all around `slog`. Didn't do any further research about potentially helpful companion packages since opening this ticket, though. 

Switching to structured logging requires a bit of a mind shift in how to phrase log messages. The above Reddit post explains it quite well. So be aware that there's probably more to this ticket than simply replacing all log statements with `slog` method calls. It's probably quite a bit of refactoring work, with no - unfortunately - no benefit to the user whatsoever. Just keep that in mind before jumping onto this. If you think this is worth you're time nonetheless, I'd be super happy the help, of course! :raised_hands:.

finnng:
Hi @muety , the PR is ready to review here https://github.com/muety/wakapi/pull/667

muety:
Reopening this issue as migration to `slog-sentry` and `slog-http` (see [here](https://github.com/muety/wakapi/pull/667#discussion_r1723857057) and above) are still pending.

muety:
@finnng You wanna take this for round two?

finnng:
> @finnng You wanna take this for round two?

Yeah, I will find the time for it. Can you assign this ticket to me?

muety:
Any updates on this? Don't feel pressure, just wanted to hear about the current status.

I plan to implement #607 some time in the next days and intend to push a big new release then. Would be cool to have the full logging refactoring included by then. But again, no stress still!

[...]

PR #667, opened 2024-08-20
#480 Migrate from logbuch to slog

finnng:
This pull request includes changes to the logging system in multiple files. The primary goal is to replace the `logbuch` logging library with the standard `log` package and the `slog` package for structured logging.
Resolves: https://github.com/muety/wakapi/issues/480

### Notes
- slogs doens't have Fatal method, but we set default `slog.SetDefault`, it overrides default log logger as well. Then `log.Fatal` also print as structured log
- I'm only replace logbuch by slog, didn't replace SentryWrapperLogger by lib slog-sentry

### Logging System Changes:
* **Imports and Initialization:**
  * Added `log` and `log/slog` imports in `config/config.go`, `config/jobqueue.go`, `config/key_utils.go`, `config/sentry.go`, and `main.go`. [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R6) [[2]](diffhunk://#diff-64e442c17a94d4ab454e39730b7c8b160abcd74f81e7a57b7c6ed249fc93e279L5-R7) [[3]](diffhunk://#diff-1223144ed79083e40909169cd1f32855d3dd6fe9972a808e0ddf4c41748b8391L4-R6) [[4]](diffhunk://#diff-bec9004b22122629034045319bae442ca8d84ce05f4b2d6e7f574e0a38697e95L4-R14) [[5]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R8)
  * Removed `logbuch` import from the same files. [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L16-R23) [[2]](diffhunk://#diff-64e442c17a94d4ab454e39730b7c8b160abcd74f81e7a57b7c6ed249fc93e279L5-R7) [[3]](diffhunk://#diff-1223144ed79083e40909169cd1f32855d3dd6fe9972a808e0ddf4c41748b8391L4-R6) [[4]](diffhunk://#diff-bec9004b22122629034045319bae442ca8d84ce05f4b2d6e7f574e0a38697e95L4-R14) [[5]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L16)

* **Function Changes:**
  * Replaced `logbuch.Fatal` with `log.Fatal` in multiple functions in `config/config.go`, `config/key_utils.go`, and `config/sentry.go`. [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L256-R262) [[2]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L279-R285) [[3]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L304-R310) [[4]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L427-R428) [[5]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L458-R459) [[6]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L478-R479) [[7]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L496-R565) [[8]](diffhunk://#diff-1223144ed79083e40909169cd1f32855d3dd6fe9972a808e0ddf4c41748b8391L18-R38) [[9]](diffhunk://#diff-bec9004b22122629034045319bae442ca8d84ce05f4b2d6e7f574e0a38697e95L138-R146)
  * Replaced `logbuch.Warn` with `slog.Warn` in `config/config.go`. [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L336-R337) [[2]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L478-R479)
  * Replaced `logbuch.Info` with `slog.Info` in `config/jobqueue.go`, `config/sentry.go`, `helpers/http.go`, and `main.go`. [[1]](diffhunk://#diff-64e442c17a94d4ab454e39730b7c8b160abcd74f81e7a57b7c6ed249fc93e279L45-R45) [[2]](diffhunk://#diff-bec9004b22122629034045319bae442ca8d84ce05f4b2d6e7f574e0a38697e95L78-R87) [[3]](diffhunk://#diff-235a8f70b95b90ec5b3a63cb0482c2c683d799b578894c40238411ae7d5f9701L28-R28) [[4]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L127-R129) [[5]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L147-R156) [[6]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L259-R255) [[7]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L330-R326) [[8]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L369-R367)
  * Updated `SentryWrapperLogger` to use `slog.Logger` instead of `logbuch.Logger` in `config/sentry.go`. [[1]](diffhunk://#diff-bec9004b22122629034045319bae442ca8d84ce05f4b2d6e7f574e0a38697e95L30-R48) [[2]](diffhunk://#diff-bec9004b22122629034045319bae442ca8d84ce05f4b2d6e7f574e0a38697e95L78-R87)

* **Miscellaneous:**
  * Removed `logbuch` dependency from `go.mod`.


main.go
```diff
@@ -256,7 +252,7 @@ func main() {
 		middleware.StripSlashes,
 		middleware.Recoverer,
 		middlewares.NewPrincipalMiddleware(),
-		middlewares.NewLoggingMiddleware(logbuch.Info, []string{
+		middlewares.NewLoggingMiddleware(slog.Info, []string{
```
muety, line 255: Could you replace [this part](https://github.com/muety/wakapi/blob/master/middlewares/logging.go#L45-L54) of the request logging middleware with structured logging as well, please? 

If we want to keep the middleware "generic" (accept arbitrary log functions), we'd probably need to pass it another parameter like `isStructured` or sth. and then have an `if-else` to log in either plain text (like now) or structured. But I'd also be fine with having the middleware specify its own logger instead of being passed a func. Whatever you think is cleaner.

routes/login.go
```diff
@@ -341,7 +341,7 @@ func (h *LoginHandler) PostResetPassword(w http.ResponseWriter, r *http.Request)
 				if err := h.mailSrvc.SendPasswordReset(user, link); err != nil {
 					conf.Log().Request(r).Error("failed to send password reset mail to %s - %v", user.ID, err)
```
muety, line 342: I'm not 100 % happy with the fact that part of our logging is now structured, while the other half remains plain text. I think we should definitely refactor the Sentry logging to be structured (via [samber/slog-sentry](https://github.com/samber/slog-sentry)) as well.

We can totally do that in a separate pass to not bloat further bloat this PR, though. Let me know if you'd be up for that refactoring as well (don't feel obligated, of course).

Also [samber/slog-http](https://github.com/samber/slog-http) seems like a super nice replacement for our custom request logging middleware ...

muety (changes requested):
Wow, man, thanks a lot! Amazing contribution! 🙌
I added a bunch of comments, but all in all, this is mostly ready to be merged.

main.go
finnng: I missed this part, let me update it. We will use a structured logger anyway, and other implementations like Zap also follow the (key: value) approach. In case we change our mind in the future, this log func still works 🥇

main.go
```diff
@@ -144,20 +140,20 @@ func main() {
...
 	// Connect to database
 	var err error
-	logbuch.Info("starting with %s database", config.Db.Dialect)
+	slog.Info("starting with database", "dialect", config.Db.Dialect)
 	db, err = gorm.Open(config.Db.GetDialector(), &gorm.Config{Logger: gormLogger}, conf.GetWakapiDBOpts(&config.Db))
 	if err != nil {
-		logbuch.Error(err.Error())
-		logbuch.Fatal("could not open database")
+		slog.Error("could not connect to database", "error", err)
+		log.Fatal("could not open database")
```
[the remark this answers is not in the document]
finnng: Or we only need `log.Fatal` since `log.Fatal` also structure, but the problem is this method still use the INFO level
image
The options here
1. We just use the log.Fatal for now
2. We write a shared function `logFatal(...interface{})` that wrap `slog.Error` and `os.Exit()`
What do you think?

routes/login.go
finnng: Yes, I will need sometime to understand that lib and integrate with it. In the mean time we can have a few changes to make the sentry log structured. I will use the JSON handler instead of TextHandler for sentry, it will printed out the structured log as well
image

The other task is refactor conf.Log().Request(r).(Error | Warn | Info...) to structured. I missed this part, thanks for pointing out, probably a find and replace mistake

main.go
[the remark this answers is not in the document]
finnng: I updated the code to only use log.Fatal in the mean time

routes/login.go
finnng: I updated the Sentry slog handler to use NewJSONHandler. I tried to make it consistent with main.go to init the TextHandler for development and JSONHandler for production but no luck, I can't read the config.isDev() in this contructor. So we use JSONHandler for both in this case.

muety: > I can't read the config.isDev()

You could do `Get().IsDev()`, even though it's not very beautiful. Preferable option would be to pass a config from the outside.

---

What I don't like about this current implementation is the fact that a new handler gets instantiated for every single logging statement, which seems inefficient (same for the `capturingWriter` btw., which isn't optimal either).

Perhaps we could instantiate a singleton `SentryLoggerWrapper` once initially and then just reuse it inside `Log()`. Three options here:

1. Instantiate it in a `init()` method. You might not be able to call `Get()` there, though.
2. Instantiate it inside `Log()`, but only once, if not yet exiting.
3. Add some custom `InitSentryLogging()` method or so, that is called once initially (e.g. from `main()` or somewhere inside `config.go`). 

Sorry for being so nitpicky here. But I think after a technical refactoring like this, code should be at least as clean as before or cleaner.

main.go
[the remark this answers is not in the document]
muety: Not entirely happy about this. Would prefer to have errors logged to `stderr` consistently. However, this is only an interim solution until we switch to the `slog-sentry` package anyway, isn't it?

muety (changes requested):
Again, thank you so much for the effort you put into this!

main.go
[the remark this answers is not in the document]
finnng: > However, this is only an interim solution until we switch to the slog-sentry package anyway, isn't it?

We will revisit this when we switch to `slog-sentry`.

For this PR. I realize we don't need another wrapper to cover the missing of `slog.Fatal`. We can use `Log().Fatal()`, it already wrapped the `slog.Error` and `os.Exit(1)`. I pushed a commit to replace all `log.Fatal` by the `Log().Fatal()`

routes/login.go
finnng: To minimize the changes. I choose the option `Instantiate it inside Log(), but only once, if not yet exiting.`

finnng:
> Again, thank you so much for the effort you put into this!

Hi @muety, I pushed an update regarding the feedback, please help to review again

PR #668, opened 2024-08-22
fix: update Dockerfile for default db.name env

yuedanlabs:
According to README doc, the environment variables for `db.name/WAKAPI_DB_NAME` should be `wakapi_db.db`, so I updated it in Dockerfile.

the code as bellow:

```diff
ENV ENVIRONMENT=prod \
    WAKAPI_DB_USER='' \
    WAKAPI_DB_PASSWORD='' \
    WAKAPI_DB_HOST='' \
-  WAKAPI_DB_NAME=/data/wakapi.db \
+  WAKAPI_DB_NAME=/data/wakapi_db.db \
    WAKAPI_PASSWORD_SALT='' \
    WAKAPI_LISTEN_IPV4='0.0.0.0' \
    WAKAPI_INSECURE_COOKIES='true' \
```


YC:
The default in the README comes from https://github.com/muety/wakapi/blob/master/config/config.go#L134. 

However, it is true that the Dockerfile overrides that, such that the DB files ends up in `/data`, and the filename is `wakapi_db.db` instead of `wakapi.db` -- which is inconsistent. 

If my understanding is correct, the suggested change will unfortunately be a breaking change for those using Docker, and not setting a custom value for `WAKAPI_DB_NAME`.

yuedanlabs:
> The default in the README comes from https://github.com/muety/wakapi/blob/master/config/config.go#L134.
> 
> However, it is true that the Dockerfile overrides that, such that the DB files ends up in `/data`, and the filename is `wakapi_db.db` instead of `wakapi.db` -- which is inconsistent.
> 
> If my understanding is correct, the suggested change will unfortunately be a breaking change for those using Docker, and not setting a custom value for `WAKAPI_DB_NAME`.

Yes, for docker users, if the environment variable `WAKAPI_DB_NAME` is not set, the default environment variable in the Dockerfile will be used, that is, the default db.name is `wakapi.db`

So I think the environment variable settings in the Dockerfile should follow the default values ​​listed in the documentation(wakapi_db.db) and the default values ​​set in config.go(wakapi_db.db), otherwise the docker user will have a different db name if the environment variable is not set.

muety:
Thanks for the contribution. I agree that it would be nicer if the Dockerfile's default value and the actual config default would match. However, I don't think it is as important as to introduce this breaking change. People who use Docker and don't have a custom `WAKAPI_DB_NAME` set will end up with a database named `wakapi.db`, but wouldn't care about what README or `config.go` say. People who override the variable for their containers do so on purpose and won't care about the defaults either.

Issue #356, opened 2022-04-17
Docker Secrets


muety:
Refactor Docker instructions and docker-compose file to read sensitive variables from Docker Secrets.

https://hub.docker.com/_/nextcloud (section "_Docker Secrets_") can be used as an examples.

muety:
Never mind, this only seems to work with Docker Swarm, which we don't want to depend on.

1fexd:
Secrets can be used with [Docker Compose](https://docs.docker.com/compose/use-secrets/), so it would be great if you could reconsider ([Vaultwarden](https://github.com/dani-garcia/vaultwarden/wiki/Configuration-overview#loading-individual-values-from-files) also supports using secrets in this way, for example)

1fexd:
I think this can be closed :smile:

PR #679, opened 2024-09-22
Docker secrets support

1fexd:
This PR implements #356 by adding new config options following the `_FILE` convention commonly used for Docker secrets for all secrets/passwords used.

I'm not a Go expert, so I tried to implement this similarly to the existing implementation of `ParseTrustReverseProxyIPs`, however I'm not sure that's the best approach.

I've included a sample compose file (based on the existing `compose.yml` file) which demostrates the functionality and can be run like this:

`docker compose --env-file secrets.env -f compose-secrets.yml up -d`


muety:
Thanks a lot for your contribution! Will do a detailed review later, but at first sight, it looks good :+1:.

The only thing that I don't really like so much is the fact that every secret-like config option is sort of duplicated now, and required a `LoadXYZ()` function. Is this the common way of implementing the use of Docker secrets in the application? Isn't there maybe a more "generic" way that doesn't require additional config options? Something like a small wrapper script maybe, that reads those files and provides the values to the application (_only_ to that specific process) via env variables again or by injecting them into the YAML config? Or would that contradict the whole point of Docker secrets in the first place?

Do you possibly have some examples of other applications that support Docker secrets? How did they implement this?

1fexd:
Looks like both MySQL and Postgres are doing it like [this](https://github.com/docker-library/mysql/blob/master/8.4/docker-entrypoint.sh#L28) (which are also mentioned in the [Docker docs](https://docs.docker.com/compose/how-tos/use-secrets/#advanced)), so I think just using the same bash function instead of implementing it in wakapi itself would be a more sensible approach

muety:
Thanks! That's pretty much what I had in mind.

Can we implement it the same way? I appreciate your effort, but would prefer to abstract the use of secrets vs. env. variables and keep it separate from the application code. What do you think?

1fexd:
I've force pushed a new version which basically does it exactly like [Postgres](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L9) - that would have been the proper way to implement this all along :D

secrets.env
```diff
@@ -0,0 +1,2 @@
+PASSWORD_SALT=RFcxgy1GSXL4hApS
```
muety, line 1: What's the point of this file?

entrypoint.sh
```diff
@@ -1,5 +1,33 @@
...
+	fi
+	export "$var"="$val"
+	unset "$fileVar"
+}
+
+file_env "WAKAPI_PASSWORD_SALT" ""
+file_env "WAKAPI_DB_PASSWORD" ""
+file_env "WAKAPI_MAIL_SMTP_PASS"
+file_env "WAKAPI_SUBSCRIPTIONS_STRIPE_SECRET_KEY"
+file_env "WAKAPI_SUBSCRIPTIONS_STRIPE_ENDPOINT_SECRET"
```
muety, line 29: What do we need these for?

secrets.env
1fexd: Not strictly necessary, just an example that can be used with the compose file in this PR to demonstrate the feature

entrypoint.sh
1fexd: To check if any of those env variables are present with the "_FILE"-suffix and load them from the given file if they are

PR #682, opened 2024-10-04
Ignore ID in incoming heartbeat requiest

v1ack:
[Browser extension](https://github.com/wakatime/browser-wakatime) sends heartbeat in such format

```json
[
  {
    "branch": "<<LAST_BRANCH>>",
    "category": "meeting",
    "entity": "https://meet.google.com",
    "id": "0014a80e-3a79-4bdd-a678-60e801ca7f58",
    "plugin": "Google Meet",
    "project": "xxx-xxxx-xxx",
    "time": "1728043575.398",
    "type": "domain",
    "userAgent": "Chrome/129.0.0.0 mac_arm64 chrome-wakatime/4.0.2"
  },
]
```

In wakapi there is 
```golang
type Heartbeat struct {
	ID              uint64     `gorm:"primary_key" hash:"ignore"`
        ...
}
```

So we need use uuid or don't accept any id from client

There is second solution in this PR

### DIFF
--- a/models/heartbeat.go
+++ b/models/heartbeat.go
@@ -11,7 +11,7 @@ import (
 )
 
 type Heartbeat struct {
-	ID              uint64     `gorm:"primary_key" hash:"ignore"`
+	ID              uint64     `json:"-" gorm:"primary_key" hash:"ignore"`
 	User            *User      `json:"-" gorm:"not null; constraint:OnUpdate:CASCADE,OnDelete:CASCADE;" hash:"ignore"`
 	UserID          string     `json:"-" gorm:"not null; index:idx_time_user; index:idx_user_project"` // idx_user_project is for quickly fetching a user's project list (settings page)
 	Entity          string     `json:"Entity" gorm:"not null"`


muety:
Thanks for the contribution! :+1: 

Can you quickly explain what problem this will be fixing? Is the browser add-on broken because Wakapi fails to unmarshal the incoming heartbeats? :thinking: 

My only fear with this is that `id` might potentially still be used somewhere downstream. I don't feel very comfortable with removing it from the JSON (request and response) entirely. A "safer" way would be to introduce two separate "view models", one for incoming heartbeats (with `id` field ignored) and one for outgoing heartbeats sent in responses with the ID included. But this would add additional complexity and boilerplate code, so I'm not very happy with that either.

Could you please double-check that removing the ID field is safe? Especially:
- Check the code base to make sure that _no_ component relies on `id` to be present for incoming heartbeats
- Have some thoughts about where `id` could potentially be required in heartbeat responses

v1ack:
> Can you quickly explain what problem this will be fixing? 

Main problem is wakatime use uuid but wakapi use uint64

> Is the browser add-on broken because Wakapi fails to unmarshal the incoming heartbeats? 🤔

Exactly

Best decision is migrate from numeric id to uuid

> Could you please double-check that removing the ID field is safe?

Im not good in golang, but I use fixed version. it looks like everything is fine)

muety:
Alright, I'll have a deeper look in the next days and merge then.

muety:
Relates to #688.

Issue #411, opened 2022-09-28
Admin CLI


muety:
Introduce a CLI for Wakapi for basic admin tasks, including:
- Managing users (create / activate / delete)
- Resetting a user's password
- Clearing caches
- ...

Requested in the context of #410.

PR #726, opened 2024-12-23
chore: add prettier

littlecxm:
add prettier for format code & check html issue
e.g
image


muety:
Thanks for your contribution! However, I'd like to advice you to check a project's [contribution guidelines](https://github.com/muety/wakapi/blob/master/CONTRIBUTING.md) before investing a lot of effort into things that possibly aren't actually of need. This potentially saves both the contributor as well as the maintainers a lot of time.

In view of your particular change: I honestly don't see a lot of benefit in adding Prettier to a (anyways very "backend-focused") project like Wakapi. While there is _some_ benefit in detecting syntax changes (like missing closing tags in the HTML or so), I'm not sure if it's worth pulling in another piece of tooling only for that purpose - especially considering that a careful developer (which I, apparently, haven't always been 😬) would already find such mistakes through the IDEs own syntax checking at dev time.

What I strongly dislike, however, is a formatting like this:

Screenshot from 2024-12-23 08-48-50

Why would you want to dedicate a whole line only to a single character? In my opinion, this just wastes a lot of screen space and makes code _much_ harder to read. For example, check the file `time-picker.tpl.html`. The file grew from ~ 35 lines to ~ 150 lines and, at the same time, become a lot less clear to read.

So in summary, I'd like to ask you to: **either** configure Prettier in a way that it mostly keeps our current formatting and only fixes syntax errors (maybe apart from minor formatting things like too many or too few whitespaces between keywords, etc.) and additionally add a short section to the readme's _Developer Notes_ with instructions how to run the tool **or** otherwise drop the PR.

Sorry for the harsh feedback! I honestly appreciate your contribution. But you should probably have asked (as mention in the guidelines) if it's something that we actually want.

littlecxm:
Based on my personal experience, using Prettier is to help avoid simple errors in team collaboration, such as unnecessary divs, or unclosed XML elements, which I mentioned above. Since HTML/JS does not have a built-in uniform formatter like Go, and there isn't good VSCode plugin support for Golang templates (apologies, I primarily use VSCode for web development).

Prettier can also be configured to fit such habits as per your requirements.
```json
{
	"htmlWhitespaceSensitivity": "ignore",
	"bracketSameLine": true
}
```

image


Currently, there are other frameworks like ESLint, but these seem a bit heavy for Wakapi. My idea is to ensure that everyone can use a unified format for coding with minimal changes. Additionally, Prettier has nice support for mainstream IDEs (https://prettier.io/docs/en/editors.html) and can also be integrated into the CI workflow.

Thank you for your sincere response, and I also appreciate the significant effort you have put into this amazing open-source project. ❤️

muety:
Can you configure Prettier to only care about actual syntax issues and ignore pretty much everything else?

littlecxm:
I can configure Prettier limit in `views` folder
In my the other error is mention from vscode itself so I disable it in `.vscode/settings.json`

```json
{
  "html.validate.scripts": false
   ...
}
```

muety:
No, what I meant is instruct Prettier to not do any formatting, but only concern about syntax errors in the HTML.

littlecxm:
> No, what I meant is instruct Prettier to not do any formatting, but only concern about syntax errors in the HTML.

image

muety:
Thanks :+1: However, Prettier reports a bunch of false positives. Check this one for example:

image

The `li` is perfectly fine, Prettier just doesn't seem to get on with the Go template placeholders.

To be honest with you, I don't think it's worth the time to proceed this further. Tweaking Prettier to get rid of those false positives will probably be quite a hassle for very little benefit in the end. Sorry for that!

littlecxm:
Yes, I think this is the right at this time. Indeed, as you said, prettier is not very suitable for projects mainly based on HTML templates. It is more suitable for SPA (single application page). 
If possible, I can also look for other areas to help with wakapi (e.g #411) 
I'll close it :)

Issue #338, opened 2022-03-16
Calendar view


alexanderadam:
Sometimes it would be very helpful to see directly which branches of which features were worked on certain days.
A calendar overview would be nice.
Ideally a 'day' could look like this:

| Percentage | Project | Branch                         |
|-------------|----------|-------------------------|
| 30%            | wakapi | `calendar_overview`    |
| 10%            | wakapi | `LDAP-integration`      |
| 60%            | proj2    | `implement_bananas` |

And it would be great if this view would be able to show at least a week at once.
Maybe even coloured by project and with a nice bar or another graph.

This way it would be much easier to see what I was working on (i.e. for standups/retros/time tracking).

Thank you so much for wakapi! :raised_hands:

muety:
Thanks a lot, I'm glad you like Wakapi! Also, thanks for these thoughts.

I agree that some sort of calendar view would be cool to have. But I'd rather limit it to aggregations on one "level" of entities, i.e. per day percentage across projects, across languages, etc. What you suggested is - if I got it correctly - an aggregation on the combination of two entity types (project and branches, in this case). That is a bit two "specific" for my taste. 

Calendar view in general would definitely be nice though. If someone has a good idea on how such might look like or could even come up with a UI design sketch, I'd highly appreciate it!

alexanderadam:
> I'd rather limit it to aggregations on one "level" of entities, i.e. per day percentage across projects, across languages, etc. What you suggested is - if I got it correctly - an aggregation on the combination of two entity types (project and branches, in this case)

Well, the idea for this feature was born because it would be easier for devs to see what they were working on (i.e. for standups/retros/time tracking).
And the project name alone is pretty unspecific. If you'd exclusively work on wakapi you wouldn't immediately know what you were working on. Only branch names in combination with project names will give that information.
Or do you have a better idea for that?

> If someone has a good idea on how such might look like or could even come up with a UI design sketch, I'd highly appreciate it!

It's probably not a _good_ design but at least it's a start. :wink: 

![Wakapi day view](https://user-images.githubusercontent.com/372620/159565922-2dbd36dd-4cbc-407f-acc4-eb0976b0684b.png)

I obviously kept it simple therefore there's much room for improvement (i.e. maybe having a bar in the background or left of percentage column? Maybe having some visual representations/icons besides or instead of the header columns?).

```html
<div class="flex flex-col space-y-2 mt-4 rounded-md border-2 border-green-700 text-gray-300 text-l bg-gray-850">
[...]

PR #732, opened 2025-01-10
Add a "Calendar view"

justin-jiajia:
image

Implement #338 .


views/summary.tpl.html
```diff
@@ -291,6 +291,20 @@ <h4 class="font-semibold text-lg text-gray-500">{{ .TotalTime | duration }}</h4>
                 </div>
             </div>
         </div>
+        
+        <div class="row-span-1 col-span-1 sm:row-span-3 sm:col-span-3 md:row-span-3 md:col-span-3 p-4 px-6 pb-10 bg-gray-850 text-gray-300 rounded-md shadow flex flex-col w-full no-break {{ if .IsProjectDetails }} hidden {{ end }}"
```
muety, line 295: Please add some margin to the card for consistency.

image

models/summary.go
```diff
@@ -77,6 +77,12 @@ type SummaryParams struct {
 	Recompute bool
 }
 
+type DailyProjectStat struct {
```
muety, line 80: Minor one: we consider a "model" something that's persisted on the database or at least part of Wakapi's core logic. The `DailyProjectStat`, however, is only another "view" to the data (in a format that's later required for rendering the chart), so I'd rather call this a "view model", analogously to, for example `ProjectsViewModel`, `LeaderboardViewModel`, etc.

services/summary.go
```diff
@@ -536,3 +537,40 @@ func (srv *SummaryService) getProjectLabelsReverseResolver(user *models.User) mo
 		return projectStrings
 	}
 }
+
+func (srv *SummaryService) GetDailyProjectStats(from, to time.Time, user *models.User) ([]models.DailyProjectStat, error) {
```
muety, line 541: Minor one: we usually return lists of pointers (e.g. `[]*models.DailyProjectStat`). While it doesn't really make any difference, I'd still prefer to stick to that here as well, just for consistency.

services/summary.go
```diff
@@ -536,3 +537,40 @@ func (srv *SummaryService) getProjectLabelsReverseResolver(user *models.User) mo
...
+			dailyStats[date] = make(map[string]time.Duration)
+		}
+		dailyStats[date][d.Project] += d.Duration
+	}
+
+	result := make([]models.DailyProjectStat, 0)
+	for dateStr, projects := range dailyStats {
+		date, _ := time.Parse("2006-01-02", dateStr)
+		for project, duration := range projects {
+			result = append(result, models.DailyProjectStat{
```
muety, line 560: Alternatively, you could follow the "functional" approach and use `slice.Map` instead, e.g. as done [here](https://github.com/muety/wakapi/blob/1d5f51dcad86976c0f02022ac79c40d1f32aeec5/routes/compat/wakatime/v1/leaders.go#L158) and in many other places throughout the code.

muety (changes requested):
Thanks a lot for this, great contribution!

Regarding the overall implementation, I'd probably suggest a slightly different approach. You current implementation comes with a few drawbacks:

- "Duplicate" logic to aggregate durations
- Inefficient, because operating on raw heartbeats for each request instead of leveraging pre-aggregated summaries
- Not respected custom mapping or project aliases

To overcome these (and thus also improve performance), I'd probably utilize the already existing summary computation logic. I'd (very roughly speaking) do approximately this:

* Use [`SplitRangeByDays`](https://github.com/muety/wakapi/blob/1d5f51dcad86976c0f02022ac79c40d1f32aeec5/utils/date.go#L40) (search code for examples), to break down the requested interval into single days
* Use `SummaryService.Aliased()` (just like in many other places throughout Wakapi) to fetch fetch summaries for each day
* Construct the `DailyProjectStat` view model from an array of summaries, i.e. introduce methods `func NewDailyProjectStats(summaries []*models.Summaries) []*DailyProjectStat { ... }` or something, e.g. similar to [here](https://github.com/muety/wakapi/blob/1d5f51dcad86976c0f02022ac79c40d1f32aeec5/models/compat/wakatime/v1/all_time.go#L30).

routes/utils/summary_utils.go
```diff
@@ -35,6 +35,7 @@ func LoadUserSummaryByParams(ss services.ISummaryService, params *models.Summary
 		return nil, err, http.StatusInternalServerError
 	}
 
+	summary.User = params.User
```
justin-jiajia, line 38: The struct summary generated by the function seemed to have a null `User` property, so I added it.

services/summary.go
```diff
@@ -220,6 +222,32 @@ func (srv *SummaryService) Summarize(from, to time.Time, user *models.User, filt
...
+	Project  string        `json:"project"`
+	Duration time.Duration `json:"duration"`
+}
+
+// get a day-by-day summary of **each** project
+func (srv *SummaryService) NewDailyProjectStats(s *models.Summary, filters *models.Filters) ([]*DailyProjectViewModel, error) {
+	dailyProjects := make([]*DailyProjectViewModel, 0)
+	intervals := utils.SplitRangeByDays(s.FromTime.T(), s.ToTime.T())
+	for _, cur := range intervals {
+		cur_summary, err := srv.Aliased(cur[0], cur[1].Add(-1*time.Second), s.User, srv.Retrieve, filters, false)
```
muety, line 236: Can you explain what's the point of `Add(-1*time.Second)`?

services/summary.go
```diff
@@ -220,6 +222,32 @@ func (srv *SummaryService) Summarize(from, to time.Time, user *models.User, filt
...
 }
 
+type DailyProjectViewModel struct {
+	Date     time.Time     `json:"date"`
+	Project  string        `json:"project"`
+	Duration time.Duration `json:"duration"`
+}
+
+// get a day-by-day summary of **each** project
+func (srv *SummaryService) NewDailyProjectStats(s *models.Summary, filters *models.Filters) ([]*DailyProjectViewModel, error) {
```
muety, line 232: Sorry for being very picky about code architecture here, but: following the notion of a "layered architecture" / adhering to the "dependency rule", the service layer shouldn't have any referenced to view models. They rather belong to what you could call "presentation layer" or so and thus only be used at the level of `routes`. 

I'd prefer to go for this implementation:
- Move the `DailyProjectViewModel` definition to `models/view`
- In that same file, add a method `NewDailyProjectStats([]*models.Summary summaries)`
- Move this function here to inside `SummaryHandler` and change it in a way that it (1) retrieves the required list of summaries from the service and (2) calls the just mentioned `NewDailyProjectStats` to have it construct the actual view model from it

So in summary:
- Service remains mostly unchanged
- Route is responsible for fetching summaries and calling to construct view model
- View model construction (from a list of summaries) is implemented next to the view model itself

Again, sorry for dragging this out so much! :-|

muety (changes requested):
Thanks for implementing my change requests! 

Do you think we should limit the calendar view component to a maximum number of days? Because when choosing a very large – if not _All Time_ – interval, the chart becomes unreadable and super inefficient (blocked my browser tab for almost 30 seconds). Would love to hear your thoughts on that.

services/summary.go
justin-jiajia: I searched the repo for the example of `SplitRangeByDays`. I got the following code snippet:

https://github.com/muety/wakapi/blob/a02c96ef5f124a063812a277a36d8e20edcb3214/routes/compat/wakatime/v1/summaries.go#L135-L150

As it includes both `SplitRangeByDays` and `summarySrvc.Aliased`, I wrote my code according to it.

I didn't know whether the `Aliased` takes account of the heartbeats exactly sent at last second or not, so I thought this line was better as the last second (or the first second on the next day) won't be calculated twice.

muety: Thanks for explaining. You should be good to go with using the _exact_ end date, summary calculation logic will take care of the rest :+1:.

justin-jiajia: Could you please tell me a bit more? What is *exact*? Should I remove the `add(-1s)` or not?

justin-jiajia:
Thank you for pointing out my problems in architecture. I've changed my codes to let them:
- fetch 1 summary for each single day using `FetchSummaryForDailyProjectStats(params *models.SummaryParams) ([]*models.Summary, error)`
- construct a `DailyProjectViewModel` from the summaries fetched below using `NewDailyProjectStats(summaries []*models.Summary) []*DailyProjectViewModel`

So: the "presentation layer", including the `router` & `handler`, present the data to the user directly without really calculating them, while the "service layer" calculates them and store them to the database.

I'm just a student and it's my first time to take part in the contribution of such a enormous project. Sorry for breaking the coding rules :-(

models/view/summary.go
```diff
@@ -14,11 +15,32 @@ type SummaryViewModel struct {
...
+type DailyProjectViewModel struct {
+	Date     time.Time     `json:"date"`
+	Project  string        `json:"project"`
+	Duration time.Duration `json:"duration"`
+}
+
+func NewDailyProjectStats(summaries []*models.Summary) []*DailyProjectViewModel {
+	dailyProjects := make([]*DailyProjectViewModel, 0)
+	for _, summary := range summaries {
+		for _, project := range summary.Projects {
```
justin-jiajia, line 33: Maybe I can copy & paste some of this to generate summary for languages, editors...

static/assets/js/summary.js
```diff
@@ -25,10 +26,11 @@ const labelContainer = document.getElementById('label-container')
 const branchContainer = document.getElementById('branch-container')
 const entityContainer = document.getElementById('entity-container')
 const categoryContainer = document.getElementById('category-container')
+const daliyContainer = document.getElementById('daily-container')
 
-const containers = [projectContainer, osContainer, editorContainer, languageContainer, machineContainer, labelContainer, branchContainer, entityContainer, categoryContainer]
-const canvases = [projectsCanvas, osCanvas, editorsCanvas, languagesCanvas, machinesCanvas, labelsCanvas, branchesCanvas, entitiesCanvas, categoriesCanvas]
-const data = [wakapiData.projects, wakapiData.operatingSystems, wakapiData.editors, wakapiData.languages, wakapiData.machines, wakapiData.labels, wakapiData.branches, wakapiData.entities, wakapiData.categories]
+const containers = [projectContainer, osContainer, editorContainer, languageContainer, machineContainer, labelContainer, branchContainer, entityContainer, categoryContainer, daliyContainer]
```
muety, line 31: Minor one: typo here

routes/summary.go
```diff
@@ -41,6 +43,19 @@ func (h *SummaryHandler) RegisterRoutes(router chi.Router) {
 	router.Mount("/summary", r)
 }
 
+func (h *SummaryHandler) FetchSummaryForDailyProjectStats(params *models.SummaryParams) ([]*models.Summary, error) {
```
muety, line 46: Minor one: make this a private method (lower-case) and move it _below_ all public methods. Also, please stick to Go code style convention of having camel-case variables (e.g. `cur_summary` -> `curSummary`).

models/view/summary.go
muety: I'd probably just stick to projects for now.

muety:
> I'm just a student and it's my first time to take part in the contribution of such a enormous project. Sorry for breaking the coding rules :-(

No worries, dude! Getting into a new code base is hard, even more so with a maintainer who is quite opinionated about architecture and code style... I'm happy to guide you the way through Wakapi's code. Perhaps you'll even become keen to contribute on a regular basis once you're familiar with it.

muety (changes requested):
Found another two issue (maybe related):
1. It looks like either the date shown below the vertical bars is wrong (off by -1 day). If I choose "Today" (today is Jan 17th), the chart will show `2025/01/16`. Similarly, if I choose "Yesterday", the chart will instead show `2025/01/15` (the day _before_ yesterday).
2. The chart seems to omit the last day of the chosen interval. If I select "Past 7 Days", I'll end up with only six vertical bars, while the last one (today) is missing.

justin-jiajia:
> Found another two issue (maybe related):
> 
> 1. It looks like either the date shown below the vertical bars is wrong (off by -1 day). If I choose "Today" (today is Jan 17th), the chart will show `2025/01/16`. Similarly, if I choose "Yesterday", the chart will instead show `2025/01/15` (the day _before_ yesterday).
> 2. The chart seems to omit the last day of the chosen interval. If I select "Past 7 Days", I'll end up with only six vertical bars, while the last one (today) is missing.
> 
> Also, can you please comment on my [comment above](https://github.com/muety/wakapi/pull/732#pullrequestreview-2551656986) regarding very large or very small intervals? I think the chart looks weird for less than three days or so and for more than 30 days (despite rendering the entire page unresponsive).

I'm sorry but I'm unable to reproduce both of these problems under CST(UTC+8). Here's my screenshot for the 'past 7 days'.

image

Could you please provide me with your timezone & `wakapiData.dailyStats` in the source of the summary?

justin-jiajia:
> Thanks for implementing my change requests!
> 
> Do you think we should limit the calendar view component to a maximum number of days? Because when choosing a very large – if not _All Time_ – interval, the chart becomes unreadable and super inefficient (blocked my browser tab for almost 30 seconds). Would love to hear your thoughts on that.
> 
> image

Yes. The problem is especially serious when the network to the database is slow. Or, in another word, server-side calculation also costs lots of time.

I've temporary limited the interval to `(3days, 31days)`. However, if the user coded only 3 days in the past 3 month, the calendar view won't show up, either. This may lead to a worse user experience. 

BTW, shall we us [Time Cartesian Axis](https://www.chartjs.org/docs/latest/axes/cartesian/time.html) for the chart?

muety (changes requested):
> Could you please provide me with your timezone & wakapiData.dailyStats in the source of the summary?

Never mind, I think it was an issue related to my test data. Tried again and seems to be working fine.

---

> However, if the user coded only 3 days in the past 3 month, the calendar view won't show up, either. 

Yes, I think that's not optimal. There should always be exactly as many vertical bars as there are days in the selected interval, otherwise users are going to be confused. To account for that, you could **(a)** either change your view model from `[{date: '' project: '', duration: 0}]` to a nested structure like `[{ date: '', projects: [ { name: '', duration: 0 } ] }]` to ensure one entry per day is present or, **(b)** fill missing dates front-end sided (probably a hassle) or, in the worst case **(c)** add a "dummy" project `null` (or so) to empty days.

---

> BTW, shall we us [Time Cartesian Axis](https://www.chartjs.org/docs/latest/axes/cartesian/time.html) for the chart?

Not very familiar with Chart.js, but sounds useful :+1: 

---

Btw., for whatever reason the activity chart isn't showing up anymore on your branch. Can you reproduce that and, if yes, fix it?

image

justin-jiajia:
> > Could you please provide me with your timezone & wakapiData.dailyStats in the source of the summary?
> 
> Never mind, I think it was an issue related to my test data. Tried again and seems to be working fine.

I tried to fix another issue caused by time zones and this problem was resolved by the way. I didn’t realize it :)

> > However, if the user coded only 3 days in the past 3 month, the calendar view won't show up, either.
> 
> Yes, I think that's not optimal. There should always be exactly as many vertical bars as there are days in the selected interval, otherwise users are going to be confused. To account for that, you could **(a)** either change your view model from `[{date: '' project: '', duration: 0}]` to a nested structure like `[{ date: '', projects: [ { name: '', duration: 0 } ] }]` to ensure one entry per day is present or, **(b)** fill missing dates front-end sided (probably a hassle) or, in the worst case **(c)** add a "dummy" project `null` (or so) to empty days.
> 
> > BTW, shall we us [Time Cartesian Axis](https://www.chartjs.org/docs/latest/axes/cartesian/time.html) for the chart?
> 
> Not very familiar with Chart.js, but sounds useful 👍

Sorry for my vague expression. I wanted to say that [Time Cartesian Axis](https://www.chartjs.org/docs/latest/axes/cartesian/time.html) does fix the missing vertical bars that were missing if the user didn’t code then.

Specifically, the axis make “Data spread according to the amount of time between data points”, providing a more simple approach of achieving the desired result. However, the time scale requires both a date library and a corresponding _adapter_ to be present (see the link to the document above), which means that an extra JavaScript dependence will be added, drawing the size of the codebase.

> Btw., for whatever reason the activity chart isn't showing up anymore on your branch. Can you reproduce that and, if yes, fix it?
> 
> image

Will try to fix if I can reproduce.

muety:
> I wanted to say that [Time Cartesian Axis](https://www.chartjs.org/docs/latest/axes/cartesian/time.html) does fix the missing vertical bars

Perfect!

justin-jiajia:
Considering the [Time Cartesian Axis](https://www.chartjs.org/docs/latest/axes/cartesian/time.html) requires 2 extra JavaScript dependence, I implement the function using your (a) plan.

> Btw., for whatever reason the activity chart isn't showing up anymore on your branch. Can you reproduce that and, if yes, fix it?

To test the function, some test data is required. However, my network connection to the production database server is extremely slow, so the svg chart isn't showing at any pages. So, BTW, is there a CLI commend to generate some fake data?

justin-jiajia:
Forget to modify the rule of "no data". Will fix tomorrow.

muety:
> So, BTW, is there a CLI commend to generate some fake data?

You can use [`sample_data.py`](https://github.com/muety/wakapi/blob/master/scripts/sample_data.py). Requires to install a bunch of Python dependencies though.

justin-jiajia:
THX

获取Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

muety:
Let me know when you're ready for another round of review :v:.

justin-jiajia:
> Btw., for whatever reason the activity chart isn't showing up anymore on your branch. Can you reproduce that and, if yes, fix it?

It was really weird. The `data` seemed to be stored to the `activityChartSvg` of the `PetiteVue` instance. However, for certain reason it was not rendered in the page. I tested even with `${ activityChartSvg }` **alongside** the svg and it wasn't rendered either.

So I came up with testing the page with only `${ activityChartSvg }` left on the page (the `mount` param were also deleted). Surprisingly, it worked! So I committed them as a workaround. I'm not sure whether or not it's a bug of `PetiteVue`.

---



It's uncomfortable for me to look at so much space in the right of the page. Could you please fix it? (I'm not familiar with the grids system frontend)

muety (changes requested):
Your HTML was messed up, meaning your opening and closing `divs` didn't match. Please revert the last commit and instead apply the changes from the attached patch.

`fiy_activity_chart.patch`

Nevertheless, things still don't look perfectly right. Besides there being no margin before your chart the logic still doesn't seem to be correct. See attached screencast.


https://github.com/user-attachments/assets/e9dc2db2-0308-4bb8-b5cb-0eb0605f634b

justin-jiajia:
Thanks for pointing out these problems, and I'll try to fix the margin and 'no data' notification a few days later after applying the patch, but I am unable to download the patch since opening the first link resulted in an error. Thanks again for your patient guidance.
________________________________

muety:
Sorry, fixed the link!

static/assets/js/summary.js
justin-jiajia, line 1: > Nevertheless, things still don't look perfectly right. Besides there being no margin before your chart the logic still doesn't seem to be correct.

It seemed that you get this problem after revert the last commit (this commit). The problem were fixed here.

justin-jiajia:
What's more wired, after applying the patch, the margin was disappeared again. It seems to me that the HTML structure was right before (I looked through the indents by my eyes lol).

static/assets/js/summary.js
muety: Then please apply the styling-related fixes from your latest commit, but fix the activity chart's HTML structure (misplaced `div`s) according to my patch.

justin-jiajia:
Patch applied. However, I don't think it's correct to add a `</div>` ahead of the daily project time's `div` (line 16 in your patch) as it closes the grid system's `div`. So, I moved the `</div>` right after the daily project time's `div`. The margin seems to be fine.

https://github.com/muety/wakapi/blob/b3061ca2bd28b6fa725a90622c6056bc9f0a7929/views/summary.tpl.html#L292-L307

**As a result, in this commit, line 305 were deleted, while the indents of the lines above were fixed.**

muety (changes requested):
Things still don't look perfectly right.

Selected "yesterday" with no data being present:
image
-> everything alright

Selected "this week" with no data being present:
image
-> Why does it show the chart _and_ the warning message? If there's no data, please don't show the chart, but instead simply show "No data" like with all the other chart (see "Categories" in the screenshot).

justin-jiajia:
> Things still don't look perfectly right.
> 
> Selected "yesterday" with no data being present: image -> everything alright
> 
> Selected "this week" with no data being present: image -> Why does it show the chart _and_ the warning message? If there's no data, please don't show the chart, but instead simply show "No data" like with all the other chart (see "Categories" in the screenshot).

Fixed.

muety:
Thanks a lot for all the time you have put into this! :raised_hands:

muety:
Actually noticed one last bug: we have a setting called "Vibrant Colors". If it's set to "disabled", all charts (except languages) should use rather mute, greenish colors. Your new chart doesn't respect this setting and remains happy and colorful all the time. It would be super cool if you could make the daily projects stats chart adhere to that setting too.

justin-jiajia:
> Actually noticed one last bug: we have a setting called "Vibrant Colors". If it's set to "disabled", all charts (except languages) should use rather mute, greenish colors. Your new chart doesn't respect this setting and remains happy and colorful all the time. It would be super cool if you could make the daily projects stats chart adhere to that setting too.

Fixed. Also thanks you for your patient guide & careful review!

I noticed that my last commit to my fork doesn't appear here. Shall I create a new PR?

PR #735, opened 2025-01-14
Update README.md

ricristian:
Update the helm repo repository url since the current one is not updated for quite some time. It includes some bugs also config map is not synchronized with current app values. 

I would recommend to wait for few days maybe the current owner of the project will accept my changes. 

In my changes besides fixes it also includes some workflows that will generate automatically new helm charts versions based on current app releases


muety:
Thanks a lot for this contribution! I'm not really familiar with K8S or Helm charts, but your changes appear sensible to me.

PR #776, opened 2025-04-13
Add support for timeline

### LINKED ISSUE #101: Timeline chart
Like the one from WakaTime or Hakatime:

justin-jiajia:
Closes #101.

preview

As this TODO says, one duration now only has 1 file that was "most prominent" (i.e. was coded on most) with the duration. So the timeline does so.

https://github.com/muety/wakapi/blob/a842467067de8199cecb74b3255526d1e2115972/services/duration.go#L224-L231

And the duration service doesn’t respect alias, does it?

BTW, it now only generates time line for less than 7 days. Maybe decrease the limit?


muety:
Thanks a lot for another great contribution! 🙌 First of all: great work! Nevertheless, a couple of thoughts on this:

## Crammed summary page
I'm a bit worried about the summary page becoming too loaded, (a) from a UX perspective and (b) in terms of performance. The "summary" has become much more than just a quick summary, and instead pretty much shows everything Wakapi _can_ possibly show, except for per-project branch- and file stats. Also, since `Duration`s are, compared to `Summary`s, on a very granular level, the number of bytes transferred over the wire for only the HTML part of initially showing "_Today_" has, in my quick test, almost tripled thanks to the timeline chart. Of course, we're still talking about < 1 MB, but performance and efficiency is one of Wakapi's primary goals.

So let me propose two different alternatives to work around this:
* Option 1 (my preference): similarly to Hakatime (see screenshot in #101), make the timeline chart always only show the last 24 hours within the selected range. I think the main purpose of this chart is to get an idea about which times of the day a user is being productive. In that case, it doesn't really matter to look super far behind. 
* Option 2: Introduce a second page (perhaps called "Productivity" or something? 🤷‍♂️) to house the "Timeline" and "Daily Timeline" charts. Haven't really thought through this, though. Perhaps you can propose some concept for "debloating" the front page by outsourcing certain functionality to somewhere else in a sensible way?

## Naming / wording
Terminology has become somewhat confusing lately. For one thing, there is the "Timeline" chart, which, in the code is referred to as "daily stats". On the other hand, we have "Daily Timeline", named "timeline" in the code. We should tidy this up. For example, consistently call the "daily stats" thing "timeline" on UI and in the code, and the new feature perhaps "hourly breakdown" or something? Happy to hear better suggestions!

## ECharts
As per our design goal of [_minimalism_](https://github.com/muety/wakapi/wiki/Contributing#minimalism) I'm not happy about pulling in another charting library. Can this functionality be realized with ChartJS instead? Otherwise, if you can give striking reasons why ECharts is a lot better, we could also refactor _everything_ else to use ECharts, but for obvious reasons I'm reluctant about that. Two parallel charting libraries is not a good option, though.   

Let me know that you think! Again, thanks a lot for your efforts!

models/view/summary.go
```diff
@@ -48,6 +60,30 @@ func NewDailyProjectStats(summaries []*models.Summary) []*DailyProjectsViewModel
 	return dailyProjects
 }
 
+func NewTimelineViewModel(durations models.Durations) []*TimelineViewModel {
+	timeline := make([]*TimelineViewModel, 0)
+	// Group by project
```
muety, line 65: Using [`GroupWith`](https://github.com/duke-git/lancet/blob/main/docs/en/api/packages/slice.md#GroupWith) for that would probably a bit more elegant and also more efficient than two nested loops, what do you think?

muety (changes requested):
Thanks a lot for another great contribution! 🙌 First of all: great work! Nevertheless, a couple of thoughts on this:

## Crammed summary page
I'm a bit worried about the summary page becoming too loaded, (a) from a UX perspective and (b) in terms of performance. The "summary" has become much more than just a quick summary, and instead pretty much shows everything Wakapi _can_ possibly show, except for per-project branch- and file stats. Also, since `Duration`s are, compared to `Summary`s, on a very granular level, the number of bytes transferred over the wire for only the HTML part of initially showing "_Today_" has, in my quick test, almost tripled thanks to the timeline chart. Of course, we're still talking about < 1 MB, but performance and efficiency is one of Wakapi's primary goals.

So let me propose two different alternatives to work around this:
* Option 1 (my preference): similarly to Hakatime (see screenshot in #101), make the timeline chart always only show the last 24 hours within the selected range. I think the main purpose of this chart is to get an idea about which times of the day a user is being productive. In that case, it doesn't really matter to look super far behind. 
* Option 2: Introduce a second page (perhaps called "Productivity" or something? 🤷‍♂️) to house the "Timeline" and "Daily Timeline" charts. Haven't really thought through this, though. Perhaps you can propose some concept for "debloating" the front page by outsourcing certain functionality to somewhere else in a sensible way?

## Naming / wording
Terminology has become somewhat confusing lately. For one thing, there is the "Timeline" chart, which, in the code is referred to as "daily stats". On the other hand, we have "Daily Timeline", named "timeline" in the code. We should tidy this up. For example, consistently call the "daily stats" thing "timeline" on UI and in the code, and the new feature perhaps "hourly breakdown" or something? Happy to hear better suggestions!

## ECharts
As per our design goal of [_minimalism_](https://github.com/muety/wakapi/wiki/Contributing#minimalism) I'm not happy about pulling in another charting library. Can this functionality be realized with ChartJS instead? Otherwise, if you can give striking reasons why ECharts is a lot better, we could also refactor _everything_ else to use ECharts, but for obvious reasons I'm reluctant about that. Two parallel charting libraries is not a good option, though.   

Let me know that you think! Again, thanks a lot for your efforts!

justin-jiajia:
> * Option 1 (my preference): similarly to Hakatime (see screenshot in [Timeline chart #101](https://github.com/muety/wakapi/issues/101)), make the timeline chart always only show the last 24 hours within the selected range. I think the main purpose of this chart is to get an idea about which times of the day a user is being productive. In that case, it doesn't really matter to look super far behind.

Done.

> Terminology has become somewhat confusing lately. For one thing, there is the "Timeline" chart, which, in the code is referred to as "daily stats". On the other hand, we have "Daily Timeline", named "timeline" in the code. We should tidy this up. For example, consistently call the "daily stats" thing "timeline" on UI and in the code, and the new feature perhaps "hourly breakdown" or something? Happy to hear better suggestions!

Done

> As per our design goal of [_minimalism_](https://github.com/muety/wakapi/wiki/Contributing#minimalism) I'm not happy about pulling in another charting library. Can this functionality be realized with ChartJS instead? Otherwise, if you can give striking reasons why ECharts is a lot better, we could also refactor everything else to use ECharts, but for obvious reasons I'm reluctant about that. Two parallel charting libraries is not a good option, though.

Done the refactoring. I used ECharts initially because of its diverse examples, including that timeline and the "zoom" plugin (you may see it at the bottom of the first image), but they became unnecessary as only a 24-hour-long breakdown is generated.

And does the duration service respect aliases, or do I need to implement it on my own? (It affects the hourly breakdown.)

refactored

muety:
Wow, that was quick! I'll have a deeper look in the next days.

> And does the duration service respect aliases, or do I need to implement it on my own?

Alias resolution is only done at the level of summaries currently, see [here](https://github.com/muety/wakapi/blob/a842467067de8199cecb74b3255526d1e2115972/services/summary.go#L57). You could implement something like [this](https://github.com/muety/wakapi/blob/a842467067de8199cecb74b3255526d1e2115972/models/summary.go#L361) analogously for durations. One step further, we might consider replacing the summary-level "aliasing" by that. But I'll have to think about that more carefully. Until then, there must remain the ability to also retrieve "raw" durations like now.

muety (changes requested):
I did a couple of small code changes and pushed them to [`justin-jiajia/timeline-chart`](https://github.com/muety/wakapi/tree/justin-jiajia/timeline-chart). Can you please rebase your branch onto that one?

Last two remarks:
- Only looking at the chart, there's currently no way to tell which date a tick on the x-axis corresponds to, which, I think, is a bit confusing. Would it be possible to show the date as well (maybe sacrifice second-level precision instead for brevity or so)?
    image
- How hard would it be to add the zooming functionality back in now after having switched to ChartJS? I kind of liked that.

justin-jiajia:
> How hard would it be to add the zooming functionality back in now after having switched to ChartJS? I kind of liked that. 

Just thought of it. 

A [double range slider](https://www.thecoderashok.com/blog/double-range-slider-with-min-max-value) would be a tricky way to implement this functionality. However, it only supports to adjust min/max at a click (you may try it in the link below). 

Another approach for "zoom" is also possible by using a [popular chartjs plugin](https://www.chartjs.org/chartjs-plugin-zoom/latest/guide/). However, it requires an extra library for gesture recognition.

muety:
Using chartjs-plugin-zoom seems like a good option to me! It would be even cooler if you could not only zoom, but also click & drag to move the chart area, but that's optional.

Regarding the additional library: I'd presume the library would remain functioning even without it, only that there's no pinch & pan support then. However, since Wakapi doesn't work too well on mobile anyway, I think that's acceptable. But otherwise I'd also be fine with pulling in Hammer.js in addition, since it's "only" an additional 8 kBytes.

muety (approved):
Awesome work! :clap: :pray:

PR #781, opened 2025-04-30
ci: improve docker build speed

shyim:
In that way we utilize the cross compiling of Go, instead that we have to rely on QEMU.
This took on my machine 23s to build for all three arches.

https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:alpine AS build-env
+FROM --platform=$BUILDPLATFORM golang:alpine AS build-env
 WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


muety:
This is super cool, thanks a lot! :raised_hands:.

Will the GitHab Actions CI workflow benefit from this as well, without any additional changes? Do we have to remove the `Set up QEMU` stage?

shyim:
Yes GitHub Action will benefit the most of that. We can keep the qemu I think is also needed as the second stage (the another FROM) does not specify the build platform

muety (approved):

PR #802, opened 2025-06-09
feat(api): return health check as JSON

kurtnettle:
Updated the `/health` endpoint to return a JSON response (`{"app": 1, "db": 1}`) instead of plain text.

With a structured JSON response, it will help users to monitor the wakapi health with ease. Like recently using uptime-kuma, I faced difficulty in creating the monitoring rule for wakapi due to the plain text with a `\n`.

### DIFF
--- a/routes/api/health.go
+++ b/routes/api/health.go
@@ -1,7 +1,7 @@
 package api
 
 import (
-	"fmt"
+	"encoding/json"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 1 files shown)


kurtnettle:
I also thought the same about this breaking change PR and I am extremely sorry for submitting PR out of the blue. 
I will remain careful in future. Thanks for being tolerant. (⌒_⌒;)

kurtnettle:
You forgot use the variable value for db status! 
https://github.com/muety/wakapi/blob/c6b8bf7b676445126c3c90478d295829df297976/routes/api/health.go#L45

muety:
> You forgot use the variable value for db status!
> 
> https://github.com/muety/wakapi/blob/c6b8bf7b676445126c3c90478d295829df297976/routes/api/health.go#L45

My bad, thanks for the hint!

Issue #476, opened 2023-03-25
Docker Image for 2.7.0


YC:
Hi @muety, the 2.7.0 action for Docker doesn't seem to have triggered. 
Edit: I have created docker-release branch with a commit to make the docker release action consistent with release.

muety:
Thanks for this! Any clues about why the action didn't trigger? Also, do you know of a convenient way to manually trigger that action retroactively, while "simulating" a release-publish event?

YC:
My guess is that it's related to `[skip ci]`. 
Now that it's consistent, hopefully it will work in the future with my change to the actions file. 

A method may be to delete the tag and retag it, but that's probably not the best idea from a traceability perspective. 
So yes, perhaps a manual push?

sdvcrx:
@muety @YC Hello! I wanted to bring to your attention that the current docker image version `2.7.0` lacks multi-platform support, which is causing issues when running it on an ARM-based machine.

<img width="442" alt="image" src="https://user-images.githubusercontent.com/5170232/227839303-df25ef0b-7208-4bbf-b26a-6f6fb2c2cbd3.png">
<img width="438" alt="image" src="https://user-images.githubusercontent.com/5170232/227839331-411250b8-de27-429e-9141-272afd11063f.png">

YC:
Thanks @sdvcrx. 
@muety Sorry that I didn't anticipate this.
I've tried `workflow_dispatch`, but it wouldn't trigger via web or api since the event was not present in 2.7.0. 
![image](https://user-images.githubusercontent.com/5828744/227919666-166edf54-5198-4179-8cce-09ccde19bb27.png)
Perhaps a minor release, if it's not too much trouble?

muety:
@sdvcrx Thanks for bringing this up!

@YC Thanks for looking into this!

I never really got in touch with multi-platform Docker images, but looks like "cross-compilation" is supported (see [here](https://docs.docker.com/build/building/multi-platform/)). Instead of creating a new release (without any actual changes contained), I'd probably rather prefer to just push manually built images for `arm64` and `armv7` as well.

I'll give it a shot and keep you posted.

muety:
Please try again!

For the record, this is the command I used: 

```bash
docker buildx build --push --platform linux/amd64,linux/arm64,linux/arm/v7 \
	-t ghcr.io/muety/wakapi:latest \
	-t ghcr.io/muety/wakapi:2.7.0 \
	-t ghcr.io/muety/wakapi:2.7 \
[...]

PR #805, opened 2025-06-13
fix(ci): ARG for TARGETOS and TARGETARCH

### LINKED ISSUE #804: wakapi fails to start in arm64 Docker image
**Describe the bug**
`wakapi` executable fails to start
```
/app/entrypoint.sh: line 32: /app/wakapi: cannot execute binary file: Exec format error
/app/entrypoint.sh: line 32: /app/wakapi: No error information
```

**System information**
* Wakapi version: 2.14
* Operating system
    * If Docker: `ghcr.io/muety/wakapi:2.14`
* Database (SQLite, MySQL, ... ?) N/A

YC:
Resolves #804

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ COPY ./go.mod ./go.sum ./
 RUN go mod download
 COPY . .
 
+ARG TARGETOS
+ARG TARGETARCH
 RUN GOOS=$TARGETOS GOARCH=$TARGETARCH CGO_ENABLED=0 go build -ldflags "-s -w" -v -o wakapi main.go
 
 WORKDIR /staging


YC:
When executed:
```
2025-06-13T09:52:15.2937778Z #23 [linux/amd64 build-env 7/9] RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags "-s -w" -v -o wakapi main.go
2025-06-13T09:52:22.5492310Z #24 [linux/amd64->arm/v7 build-env 7/9] RUN GOOS=linux GOARCH=arm CGO_ENABLED=0 go build -ldflags "-s -w" -v -o wakapi main.go
2025-06-13T09:52:25.1917683Z #25 [linux/amd64->arm64 build-env 7/9] RUN GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go build -ldflags "-s -w" -v -o wakapi main.go
```

muety:
I'd love to have [`workflow_dispatch`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch) events as a second trigger for the Docker workflow, combined with the ability explicitly pass in a tag number as an input. 

If you ever find yourself bored, would you mind adding this? 😄

YC:
From a build traceability perspective, this is perhaps not best practice given that the code which is built will not match the git tag, and the previous tag on image registry is replaced. 
However, I think it would be appropriate if it becomes possible to manually trigger a build off of a tag, with the code from that tag without creating a GitHub release, keeping in mind #476. 
Then, if such an issue happens again, the solution would be to create a new tag containing the fix, and manually trigger a run of that tag. What do you think?

muety:
> From a build traceability perspective, this is perhaps not best practice given that the code which is built will not match the git tag, and the previous tag on image registry is replaced.

Agree! 

> [...] manually trigger a build off of a tag, with the code from that tag without creating a GitHub release

The only thing that I'm a bit unhappy about: for me, every tag should represent a new version of the software, while fixing CI is not actually a new version in a strict sense. But I think it's fine. Can you check if we could, technically, also have Docker tags deviating from the standard semantic version format (e.g. have `2.14.0-hotfix1` or sth.)?

PR #829, opened 2025-08-14
refactor: replace colors with css variables

k2on:
This PR replaced hard coded colors with css variables to make development easier and support different themes.

Let me know if this is something that you would consider merging.


muety:
Sounds great, let's proceed that way 👍.

k2on:
I started school this week, so I am going to try to get this done this week or weekend.

k2on:
@muety I did my best to refactor the colors. I simplified them a little bit (i.e: secondary and muted). For example on the login screen, these text colors can probably be the same color.

image

Let me know what needs changing. Thanks for helping me out with this.

k2on:
Here is what `border-focused` looks like.

image

I think we should go with `muted/50`. This will get you the same color as `gray-700`.

k2on:
All done. Let me know if you see anything else.

muety:
Now live at wakapi.dev :raised_hands:.

PR #831, opened 2025-08-18
feat: add user preference for start of week

NazmusSayad:
In the original WakaTime, there is an option to set the start of the week, but that option is missing here.

---

## This PR implements the feature to customize the start of the week
image
image

### DIFF
--- a/migrations/20250818_add_start_of_week.go
+++ b/migrations/20250818_add_start_of_week.go
@@ -1,0 +1,41 @@
+package migrations
+
+import (
+	"github.com/muety/wakapi/config"
+	"gorm.io/gorm"
+)
+
+func init() {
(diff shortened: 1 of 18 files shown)


migrations/20250818_add_start_of_week.go
```diff
@@ -0,0 +1,41 @@
+package migrations
```
muety, line 1: You're actually not gonna need this migration, thanks to GORM's [auto-migrate](https://gorm.io/docs/migration.html#Auto-Migration) feature. Will remove it when merging.

muety (approved):
Wow, amazing contribution, thanks a lot! :pray: Welcome to the team :v:.

PR #859, opened 2025-10-24
feat: add support for multiple API keys in addition to the user key

### LINKED ISSUE #359: Support for multiple API tokens
I think it would be helpful to create multiple API tokens and to be able to revoke them individual.

In the current situation you need to regenerate the api token and change them on every active system.

monomarh:
Closes #359 — Support for multiple API tokens. 
Also includes enhancement requested by @wfjsw — Read-only token support for shared data access.

### DIFF
--- a/middlewares/authenticate.go
+++ b/middlewares/authenticate.go
@@ -67,8 +67,8 @@ func (m *AuthenticateMiddleware) WithRedirectErrorMessage(message string) *Authe
 	return m
 }
 
-func (m *AuthenticateMiddleware) WithFullAccessOnly(readOnly bool) *AuthenticateMiddleware {
-	m.requireFullAccessKey = readOnly
+func (m *AuthenticateMiddleware) WithFullAccessOnly(requireFullAccess bool) *AuthenticateMiddleware {
+	m.requireFullAccessKey = requireFullAccess
 	return m
 }
 
--- a/testing/wakapi_api_tests/Auth/Authenticate (header, full access key).bru
+++ b/testing/wakapi_api_tests/Auth/Authenticate (header, full access key).bru
@@ -1,5 +1,5 @@
 meta {
-  name: Authenticate (header, full access)
+  name: Authenticate (header, full access key)
   type: http
   seq: 7
 }
--- a/testing/wakapi_api_tests/Auth/Authenticate (header, readonly key).bru
+++ b/testing/wakapi_api_tests/Auth/Authenticate (header, readonly key).bru
@@ -1,5 +1,5 @@
 meta {
-  name: Authenticate (header, readonly)
+  name: Authenticate (header, readonly key)
   type: http
   seq: 8
 }


config/config.go
```diff
@@ -544,6 +543,7 @@ func Load(configFlag string, version string) *Config {
 	config.InstanceId = uuid.Must(uuid.NewV4()).String()
 	config.App.Colors = readColors()
 	config.Db.Dialect = resolveDbDialect(config.Db.Type)
+	slog.Info("loaded configuration", "environment", config.Env, "version", config.Version, "db_dialect", config.Db.Type)
```
muety: Why put this log line just exactly here, in the middle of config loading? Why not at the end?

middlewares/authenticate.go
```diff
@@ -65,6 +67,11 @@ func (m *AuthenticateMiddleware) WithRedirectErrorMessage(message string) *Authe
 	return m
 }
 
+func (m *AuthenticateMiddleware) WithOnlyRWApiKeys(onlyRW bool) *AuthenticateMiddleware {
+	m.onlyRWApiKeys = onlyRW
```
muety: This attribute is not actually being checked somewhere, is it?

services/user.go
```diff
@@ -101,17 +104,48 @@ func (srv *UserService) GetUserByKey(key string) (*models.User, error) {
 		return nil, errors.New("key must not be empty")
 	}
 
+	// TODO: does it make sense to use cache by api key if it's not being used?
```
muety: Fair point. Please delete these lines.

Dockerfile
```diff
@@ -1,26 +1,33 @@
 FROM --platform=$BUILDPLATFORM golang:alpine AS build-env
 WORKDIR /src
 
-RUN wget "https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh" -O wait-for-it.sh && \
-    chmod +x wait-for-it.sh
+RUN wget "https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh" -O /bin/wait-for-it.sh && \
```
muety: Please explain.

models/api_key.go
```diff
@@ -0,0 +1,14 @@
+package models
+
+type ApiKey struct {
```
muety: We now have API keys stored in two different places. There is the "main" API key as part of the user model and additional API keys in form of this many-to-many relation. I think that's somewhat confusing both for a user and for us as developers. What do you think?

muety (changes requested):
Thanks a lot for this contribution! Looks like it's still very much in progress, though?

Apart from that, I'd really appreciate contributors to first have a look at our [guidelines](https://github.com/muety/wakapi/wiki/Contributing) before pushing a PR. It would be great if people first double-checked on an issue for whether they may go about implementing it before jumping straight in.

muety:
Any updates on this?

monomarh:
> Any updates on this?

Hey man, it's been a tough week :) I fixed all your comments and removed all the junk code I added to enable hot reload and dlv. Thank you for having the patience to look through it all. I didn't notice that VS code added it all to the index itself. By the way, I haven't programmed on GO before, so I would really appreciate any comments you may have!

muety:
Thanks for implementing all those changes! And sorry that it took me another week to get back to this now. I think it's a great contribution, especially considering that it's your first time contributing to a Go project :+1:.

The only one last thing I'd like to ask you for is to add some unit tests for this new functionality, i.e. primarily extend the authentication middleware's set of test cases. Add positive and negative test cases for authenticating with a "non-main" API key and please also add a test that validates the rw- vs read-only logic. 

After that's done, the PR is ready to be merged.

middlewares/authenticate.go
```diff
@@ -65,6 +67,11 @@ func (m *AuthenticateMiddleware) WithRedirectErrorMessage(message string) *Authe
 	return m
 }
 
+func (m *AuthenticateMiddleware) WithFullAccessOnly(readOnly bool) *AuthenticateMiddleware {
```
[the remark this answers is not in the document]
monomarh: thanks, updated

PR #879, opened 2025-11-28
feat: allow to disable local and OIDC signup separately. Closes #868

### LINKED ISSUE #868: Ability to disable local signup and OIDC separately
I would like to disable global sign up but still allow my OIDC provider(s) to sign up users.

My use-case is that I want to use authentik without trusted headers but still be able to register my authentik users without allowing them to register via Wakapi.

vanutp:
Not sure if it's okay to enable it by default, given that there might be users that don't want this and have disabled the regular allow_signup setting for this reason.


README.md
```diff
@@ -189,6 +189,7 @@ You can specify configuration options either via a config file (default: `config
...
 | `security.cookie_max_age` /<br> `WAKAPI_COOKIE_MAX_AGE`                                     | `172800`                                         | Lifetime of authentication cookies in seconds or `0` to use [Session](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Define_the_lifetime_of_a_cookie) cookies        |
 | `security.allow_signup` /<br> `WAKAPI_ALLOW_SIGNUP`                                         | `true`                                           | Whether to enable user registration                                                                                                                                             |
+| `security.oidc_allow_signup` /<br> `WAKAPI_OIDC_ALLOW_SIGNUP`                               | `true`                                           | Whether to enable user registration                                                                                                                                             |
```
muety: Please make description be clear about the fact that this is enables registration specifically for OIDC.

PR #904, opened 2026-01-30
Get username from custom claim and request additional scopes

Schlauer-Hax:


config/oidc.go
```diff
@@ -51,6 +61,29 @@ func (token *IdTokenPayload) Username() string {
 	return ""
 }
 
+func (token *IdTokenPayload) getClaimValue(claimName string) string {
+	switch claimName {
+	case "preferred_username":
```
muety: Perhaps use reflection to retrieve these keys "dynamically"? I have a fear that it this will be somewhat error-prone in the future, as there's a chance of simply forgetting to adapt this method when modifying the struct. Up to you though.

Schlauer-Hax: Now that you mention it, the switch section is not needed at all since we store all claims in token.CustomClaims and i've added the reflection part with token.CustomClaims below already, right?

muety:
> We also have a private instance setup that uses kanidm [...]

_kanidm_ looks super promising and like a lightweight alternative to KeyCloak, Authentik & Co. Would you, by any chance, be willing to contribute a short "how to" section for our [wiki](https://github.com/muety/wakapi/wiki/OpenID-Connect-login-(SSO)#providers) on how to set up Wakapi with _kanidm_? :innocent: 

Apart from that, please give me a ping once this is ready for review again.

PR #908, opened 2026-02-07
Add search functionality to Projects page

### LINKED ISSUE #905: Search bar in Projects page
As title says, maybe add a search bar to the Projects page? Sorry, I just have a lot of projects.
Currently, I use the project filter dropdown in Dashboard and type the first letter to "search". It's a viable alternative, but not a great solution. Also, this means I can't see the color intensity like I do in the Projects page!

seriaati:
Resolves #905

Tested locally with my own database (SQLite).

Screenshot:
image

It is implemented by modifying the backend to support passing in a 'q' parameter (like `/projects?q=hoyo`) and adding a search bar in the frontend.

This is my first time contributing to a Go project, so if anything can be implemented in a more optimal way (especially the SQL query part), please let me know.

### DIFF
--- a/routes/projects.go
+++ b/routes/projects.go
@@ -57,6 +57,9 @@ func (h *ProjectsHandler) buildViewModel(r *http.Request, w http.ResponseWriter)
 	}
 
 	pageParams := utils.ParsePageParamsWithDefault(r, 1, 24)
+	// note: pagination is not fully implemented, yet
+	// count function to get total item / total pages is missing
+	// and according ui (+ optionally search bar) is missing, too
 	query := strings.TrimSpace(r.URL.Query().Get("q"))
 
 	var err error


services/heartbeat.go
```diff
@@ -285,18 +286,25 @@ func (srv *HeartbeatService) GetUserProjectStats(user *models.User, from, to tim
...
+	if search != "" {
+		h := sha256.Sum256([]byte(strings.ToLower(strings.TrimSpace(search))))
+		searchHash = fmt.Sprintf("%x", h[:4])
+	}
+	cacheKey := fmt.Sprintf("project_stats_%s_%d_%d_%d_%d_%s", user.ID, from.Unix(), to.Unix(), limit, offset, searchHash)
 	if results, found := srv.cache.Get(cacheKey); found && !skipCache {
 		return results.([]*models.ProjectStats), nil
-	} else if results, found := srv.cache.Get(fmt.Sprintf("project_stats_%s_%d_%d_%d_%d", user.ID, from.Unix(), to.Unix(), math.MaxInt32, 0)); found && !skipCache {
-		return utils.SubSlice[*models.ProjectStats](results.([]*models.ProjectStats), uint(offset), uint(offset+limit)), nil
+	} else if search == "" {
```
muety: Why separate treatmeant of `search == ""` and `search != ""` when probing the cache?

repositories/heartbeat.go
```diff
@@ -270,83 +270,69 @@ func (r *HeartbeatRepository) GetUserProjectStats(user *models.User, from, to ti
...
-				ps.cnt as count,
-				ls.language as top_language
-			from project_stats ps
-					 left join language_stats ls on ps.project = ls.project and ls.rn = 1
-			order by ps.last desc
-	`
+	searchClause := ""
+	if search != "" {
+		args = append(args, sql.Named("search", "%"+search+"%"))
+		if r.config.Db.IsPostgres() {
```
muety: Minor: use `condition.Ternary` here (see below).

routes/projects.go
```diff
@@ -56,14 +57,12 @@ func (h *ProjectsHandler) buildViewModel(r *http.Request, w http.ResponseWriter)
 	}
 
 	pageParams := utils.ParsePageParamsWithDefault(r, 1, 24)
-	// note: pagination is not fully implemented, yet
-	// count function to get total item / total pages is missing
-	// and according ui (+ optionally search bar) is missing, too
+	query := strings.TrimSpace(r.URL.Query().Get("q"))
```
muety: Any specific reason you removed my comments there?

muety (changes requested):
Thanks for your contrib! Just some minor remarks...

services/heartbeat.go
```diff
@@ -285,18 +286,25 @@ func (srv *HeartbeatService) GetUserProjectStats(user *models.User, from, to tim
 		offset = pageParams.Offset()
 	}
 
-	cacheKey := fmt.Sprintf("project_stats_%s_%d_%d_%d_%d", user.ID, from.Unix(), to.Unix(), limit, offset)
+	searchHash := ""
+	if search != "" {
+		h := sha256.Sum256([]byte(strings.ToLower(strings.TrimSpace(search))))
```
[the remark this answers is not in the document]
seriaati: Because I thought using the raw search string in the cache key could create very large keys, but now I that I think about it, this shouldn't be a problem.

Issue #944, opened 2026-06-12
Proper handling of OIDC


tooboredtocode:
Currently OIDC tokens are stored and checked in addition to the auth cookie.

This is not how OIDC tokens are supposed to work though. OIDC should work in one of the two ways:
1. OIDC is simply used for identifying the user and then the application takes over the session management
2. The application uses the session management of the OIDC provider

Currently neither is implemented properly.

### Proper Implementation of Option 1
To implement Option 1 the OIDC access token should not be stored and thus the validity/expiration should not be checked.

A PR implementing this can be found here: #945 

<details>
<summary>
This would mean the following lines need to be removed:
</summary>

https://github.com/muety/wakapi/blob/26c2e4e0fa487dfd5c0ef57047294560983c6ec6/routes/login.go#L417-L418
https://github.com/muety/wakapi/blob/26c2e4e0fa487dfd5c0ef57047294560983c6ec6/routes/login.go#L499
https://github.com/muety/wakapi/blob/26c2e4e0fa487dfd5c0ef57047294560983c6ec6/middlewares/authenticate.go#L84-L88
https://github.com/muety/wakapi/blob/26c2e4e0fa487dfd5c0ef57047294560983c6ec6/middlewares/authenticate.go#L224-L250

</details>

### Proper Implementation of Option 2
To implement Option 2 users authenticated with OIDC should not receive an authentication cookie. Instead they receive the OIDC access token and importantly the refresh token as well. (How they will be stored is an implementation detail,  but separate cookies for both could work, importantly both do not need to be stored in encrypted/authenticated cookies)

Now to authenticate a logged in user the auth should work like this:
- Check if the auth_cookie is set
  - if so use the existing flow
- Check if the OIDC cookies are present
- Get the access token and validate it
  - if its valid and not expired, continue
  - if its expired, try to get another token with the refresh token
    - if that succeeds, set the cookie and continue
  - redirect to the authentication provider
- fetch the user using the access token (ie via the subject field)

A PR implementing this can be found here: #947

muety:
Thanks for bringing this up. Option 1 (session management done by Wakapi itself) was indeed our intended way to implement this.

May I ask, how did you come across this "inconsistency" / how did you get aware of this?

tooboredtocode:
I set up my wakapi instance to authenticate with Cloudflare Auth, which I use for pretty much all of my applications. They default to rather short id_tokens which led to me getting logged out quite often, so I went to inspect the cause.

[...]

PR #947, opened 2026-06-13
fix: proper handling of oidc

tooboredtocode:
Implement Option 2 of #944 (alternative PR to #945)


muety:
Closing in factor of #945 (will give that one an in-depth review later).

muety:
Actually, after I contemplated about this for another bit, I think this option is actually the cleaner one, yet a little more complex. Reopening the PR. Review to follow soon.

config/config.go
```diff
@@ -46,12 +46,14 @@ const (
...
 
-	CookieKeySession               = "wakapi_session"
-	CookieKeyAuth                  = "wakapi_auth"
-	SessionValueOidcState          = "oidc_state"
-	SessionValueOidcIdTokenPayload = "oidc_id_token"
-	SessionValueWebAuthn           = "webauthn_session"
-	SessionValueWebAuthnExpiresAt  = "webauthn_session_expires_at"
+	CookieKeySession              = "wakapi_session"
+	CookieKeyAuth                 = "wakapi_auth"
+	CookieKeyOidcIdToken          = "oidc_id_token"
```
muety, line 52: Even though there's no "objectively" valid reason for this: my personal preference would be to have just _one_ cookie for the OIDC stuff that encode all three values. If not too much of a hassle, would you mind changing this?

middlewares/authenticate_test.go
```diff
@@ -283,107 +278,4 @@ func TestAuthenticateMiddleware_tryGetUserByTrustedHeader_Signup(t *testing.T) {
 	}
 }
```
muety, line 285: Can we have some basic tests for an authentication flow that involves `tryGetUserByOidc()` (besides the _route_-level tests)?

routes/utils/oidc.go
```diff
@@ -94,3 +59,54 @@ func DecodeOidcIdToken(token string, provider *conf.OidcProvider, ctx context.Co
 	payload.UsernameClaim = provider.UsernameClaim
 	return &payload, nil
 }
+
+func RefreshOidcIdToken(ctx context.Context, provider *conf.OidcProvider, refreshToken string) (*oauth2.Token, error) {
+	ts := provider.OAuth2.TokenSource(ctx, &oauth2.Token{RefreshToken: refreshToken})
+	return ts.Token()
+}
+
+func ExtractOidcAuth(w http.ResponseWriter, r *http.Request, config *conf.Config) (*conf.IdTokenPayload, error) {
```
muety: Minor one: passing the `Config` as an argument here feels a bit like a code smell. If you can, spontaneously, think of a somewhat cleaner way please go for it. Otherwise, it's not really a big deal either.

muety (changes requested):
Thanks a lot for you contributions!

routes/utils/oidc.go
tooboredtocode: Theoretically we could throw an id_token expired error, and have the caller refresh the token.
Imho doing it this way is cleaner though, since we we need the oidc provider to refresh the token, so passing the config is unavoidable.

config/config.go
tooboredtocode: Fair enough, that is no problem, though how would you encode it? Json and then base64 or some other way?

middlewares/authenticate_test.go
tooboredtocode: This is my first time writing go code, so i am not that familiar with go test, but I might be able to do something today or tomorrow...

routes/utils/oidc.go
muety: Nah, I think it's cleaner to have it part of the `ExtractOidcAuth` function. But I'd probably opt for the function retrieving the config itself via `conf.Get()` instead of have it be passed as an argument.

config/config.go
muety: Yes, that's what I was thinking of. But actually, never mind. Just leave it this way, should be fine like that :+1:.

routes/utils/oidc.go
tooboredtocode: Oh lol, didn't see that function, yeah that'd probably be cleaner...

tooboredtocode:
@muety I added the tests and additionally removed the "offline_access" scope from the scopes automatically added. This means that admins wanting to use refresh tokens, need to manually add this back in.

I decided to do it this way, since some oidc providers to not support this scope and those would not be able to be used otherwise.

muety:
Awesome! One last remark / change request:

We currently store the ID- and refresh token cookies in plain text. Since especially the refresh token is quite sensitive data, I don't think that's a good practice. Can you please change your implementation to use `securecookie` to encrypt those cookies, just as already being done for `wakapi_auth`? In the course of that, you'll probably also want to consolidate the three individual cookies into one (see our [previous discusion](https://github.com/muety/wakapi/pull/947/changes#r3409368434) on that). So in summary I'd propose to:

1. Create a minimal JSON struct containing all of the current `oidc_id_token`, `oidc_refresh_token` and `oidc_provider`
2. Encrypt that entire struct using the `securecookie` mechanism (which uses the new `session_key`)
3. Store that single, encrypted string in a unified `wakapi_oidc_session` cookie

Also, I added a bit of documentation for the newly introduced config options. You may just apply this patch as a new commit.

[add-docs-for-session-key.patch](https://github.com/user-attachments/files/29042925/add-docs-for-session-key.patch)

Sorry for the back and forth. Thanks a lot for your contribution though!

tooboredtocode:
> Since especially the refresh token is quite sensitive data, I don't think that's a good practice.

As mentioned in the issue, both of those are perfectly fine to store in plain text, since they are only shared with the client. As with all the other cookies, they are http only so xss is not a problem.

An attacker maliciously modifying them is also not an issue, since the id_token is a jwt and thus signed and already verifiable. The refresh token (at least from our perspective) is a random string that the oidc server black box verifies so there is no need to protect it even further.

All in all further encrypting them would yield no added security, at the cost of added complexity and computation.

muety:
Fair point. Did another bit of research and even though it still feels somewhat odd, I think it's save to store ID- and refresh tokens as `HttpOnly` cookies. [Here's](https://stackoverflow.com/questions/57650692/where-to-store-the-refresh-token-on-the-client) an interesting discussion on the topic.

PR #950, opened 2026-06-18
fix: relay middleware no longer eats local writes

alexanderadam:
If you have Wakapi configured to relay heartbeats to WakaTime (or another Wakapi instance), the relay middleware was quietly eating some of your own heartbeats before they reached the local database.

The local store handler runs right after the relay middleware and the relay middleware was rewriting `r.Body` to hold only the "new" heartbeats (the ones not yet seen, sent to avoid loops).

The store handler then read that trimmed body and saved only those.
Anything already relayed at least once, was forwarded upstream but never written locally.
From the user's side it looks like random missing minutes in the dashboard.

Now `filterByCache` returns the filtered bytes instead of mutating `r.Body`. The upstream send uses those and the local store sees the full original payload.
This way everybody gets to keep their heartbeats.

### DIFF
--- a/middlewares/custom/wakatime.go
+++ b/middlewares/custom/wakatime.go
@@ -21,6 +21,9 @@ import (
 
 const maxFailuresPerDay = 100
 
+// not really an error. it's the "seen all of these already, go back to sleep" signal.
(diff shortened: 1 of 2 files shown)


muety:
Good catch, thanks! :+1:

Issue #952, opened 2026-06-22
Hourly Breakdown chart permanently shows "No data" for active users (hard 200-segment cap)


NuclleaR:
The **Hourly Breakdown** chart on the summary page silently renders "No data" for any
interval whose 24h window contains more than 200 duration segments — even when there is
plenty of activity and every other widget on the page is populated.

In `routes/summary.go`, the view model is only built when the segment count is `<= 200`:

```go
if durations, err := h.durationSrvc.Get(...); err == nil {
    // for excessively many small segments, plotting is too performance-heavy ...
    if len(durations) <= 200 {
        hourlyBreakdown = view.NewHourlyBreakdownViewModel(...)
    }
}
```
If the threshold is exceeded, hourlyBreakdown stays empty and the frontend falls back to
the "No data" placeholder. There is no log line or UI hint that the chart was disabled
rather than empty, so it looks like a bug/data loss.

### To Reproduce
Have a day with many short, fragmented segments (easy to hit when using AI coding agents / editors that rapidly switch files, branches, etc.).
Open Summary → interval=today.
Every chart shows data, but Hourly Breakdown shows "No data".
Why segment counts explode
Durations are split on every change of the group hash (project, entity, branch,
language, editor, category, …). Tooling that touches many files in quick succession
(AI agents, formatters, etc.) easily produces several hundred sub-minute segments per day.
In my case a normal ~4.5h day produced 396 segments, so the chart was never shown.

Inconsistent client metadata makes it worse (e.g. category reported as both "AI Coding"
and "ai coding", or editor as "Claude code"/"Opus"/"VS Code" for the same work),
since each variation starts a new segment — but that's a separate, client-side issue.

### Expected behavior
The chart should remain readable (and visible) regardless of how fragmented the underlying
heartbeats are — similar to WakaTime's per-project "Today" timeline, which coalesces
contiguous activity into solid blocks instead of plotting every micro-segment.

### Proposed fix
Coalesce chronologically-contiguous segments of the same project (gap smaller than the
user's heartbeat timeout) into single blocks inside NewHourlyBreakdownViewModel, before
they reach the chart. This:

- collapses hundreds of slivers into a handful of readable blocks per project;
- removes the need for the hard <= 200 cap entirely (addresses the original browser-freeze concern from #871 at its root, since the bar count is now bounded by activity blocks, not raw segment count);
- lets the projects be ordered deterministically (e.g. by total activity) instead of the current random map-iteration order on the Y-axis.

[...]

PR #961, opened 2026-07-24
fix: coalesce hourly breakdown segments to fix silent 'No data' chart (#952)

AdarshJ173:
## Problem

The **Hourly Breakdown** chart on the summary page silently shows "No data" whenever a 24h window produces more than 200 duration segments. This is common with AI-assisted editors (Claude Code, Copilot, etc.) that rapidly switch files and generate hundreds of sub-minute segments per day — see #952.

The root cause was a hard `<= 200` guard in `routes/summary.go` that discarded the entire view-model rather than addressing the browser-freeze concern at its source.

## Solution

Coalesce chronologically contiguous segments of the same project inside `NewHourlyBreakdownViewModel`, **before** they reach the chart. Two adjacent segments are merged when the gap between them is ≤ the user's per-record `Timeout` (heartbeat gap preference, stored on each `Duration`). This bounds the bar count to the number of real activity blocks, not raw segment count.

### `models/view/summary.go`
- Add `Timeout time.Duration` field to `HourlyBreakdownItem` (populated from `duration.Timeout`, so per-user preferences are respected).
- Add `TimeEnd()` helper on `HourlyBreakdownItem`.
- Add `coalesceHourlyBreakdownItems(items HourlyBreakdownItems) HourlyBreakdownItems` — a single linear pass that merges adjacent same-project items whose gap ≤ `Timeout`.
- Refactor `NewHourlyBreakdownViewModel`: inline the sort (was a separate `slice.Map`) and call `coalesceHourlyBreakdownItems` per project immediately after sorting.

### `routes/summary.go`
- Remove the `if len(durations) <= 200` guard. Coalescing in the view layer means the chart will always have a manageable number of bars regardless of raw segment count.

## Testing

Manual: a user with >200 segments/day (AI coding session) will now see a populated, readable chart instead of "No data".

Unit test coverage: the existing `NewHourlyBreakdownViewModel` call path is exercised through the existing service-level tests. A targeted unit test for `coalesceHourlyBreakdownItems` covering contiguous-merge, gap-boundary and empty-input cases can be added — happy to include if the maintainer prefers.

---

> This contribution was developed with AI assistance (Perplexity).

### DIFF
--- a/models/view/summary.go
+++ b/models/view/summary.go
@@ -57,6 +57,15 @@ type HourlyBreakdownItem struct {
 	Duration time.Duration `json:"duration"`
 	Entity   string        `json:"entity"`
 	Project  string        `json:"-"`
+	// Timeout is the heartbeat gap threshold used to decide whether two
+	// adjacent segments belong to the same activity block during coalescing.
+	// It mirrors the per-user models.Duration.Timeout value.
+	Timeout time.Duration `json:"-"`
+}
+
+// TimeEnd returns the exclusive end time of this item.
+func (h *HourlyBreakdownItem) TimeEnd() time.Time {
+	return h.FromTime.Add(h.Duration)
 }
 
 func NewTimelineViewModel(summaries []*models.Summary) []*TimelineViewModel {
@@ -83,6 +92,7 @@ func NewHourlyBreakdownItems(durations models.Durations, resolve models.AliasRes
 				Duration: duration.Duration,
 				Entity:   duration.Entity,
 				Project:  duration.Project,
+				Timeout:  duration.Timeout,
 			}
 		})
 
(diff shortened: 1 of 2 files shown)


muety (changes requested):
Thanks for this contrib, and sorry for my review taking so long. I pushed a view minor changes to [952-coalesce-duration-segments](https://github.com/muety/wakapi/tree/952-coalesce-duration-segments), so please rebase your branch onto that one.

Two additional change requests though: 
1. Even after coalescing, if the heartbeats are super granular (user codes in many different projects "simultaneously"), we'll still end up with many small segments, potentially crashing the UI. So with your coalescing logic, we didn't actually _solve_ that original issue, only made it less likely to manifest. I think we should still keep a hard limit on the (now coalesced) number of durations. I see three options, let me your thoughts on them:
    1. **option: hard-limit:** same as before, i.e. don't render the chart if > 200 (coalesced) segments
    2. **option: truncate:** simply "cut" after ~ 200 segments, yielding an "incomplete" chart
    3. **option: coarser aggregation:** run the coalescing again, only with a larger _timeout_ now. Didn't really think through this properly though, so not sure if this is any sensible at all.
2. Please add some unit tests for the new coalescing logic.

I'm curious to hear your thoughts on (1).

muety:
What are your thoughts on this, @AdarshJ173?

